PR #4, opened 2021-09-07
Yearn Finance suite module.

mauricio1802:
This PR add a `npm module` with multiple agents for yearn finance.


e1Ru1o (commented):
I have some comments on the current approach that I think is better to discuss in slack

PR #8, opened 2021-09-13
Reentrancy counter agent

e1Ru1o:
Agent that detect reentracy calls using the call stack provided in the tx traces

### DIFF
--- a/reentrancy-counter/README.md
+++ b/reentrancy-counter/README.md
@@ -1,0 +1,27 @@
+# High Gas Agent
+
+## Description
+
+This agent detects transactions with high gas consumption
+
+## Supported Chains
+
+- Ethereum
+- List any other chains this agent can support e.g. BSC
+
+## Alerts
+
+Describe each of the type of alerts fired by this agent
+
+- FORTA-1
+  - Fired when a transaction consumes more gas than 1,000,000 gas
+  - Severity is always set to "medium" (mention any conditions where it could be something else)
+  - Type is always set to "suspicious" (mention any conditions where it could be something else)
+  - Mention any other type of metadata fields included with this alert
+
+## Test Data
+
+The agent behaviour can be verified with the following transactions:
+
+- 0x1b71dcc24657989f920d627c7768f545d70fcb861c9a05824f7f5d056968aeee (1,094,700 gas)
+- 0x8df0579bf65e859f87c45b485b8f1879c56bc818043c3a0d6870c410b5013266 (2,348,226 gas)
--- a/reentrancy-counter/README.md
+++ b/reentrancy-counter/README.md
@@ -1,27 +1,20 @@
-# High Gas Agent
+# Reentrancy counter
 
 ## Description
 
-This agent detects transactions with high gas consumption
+This agent detects report transaction where reentrancy ocur, based on the calls stack provided in the transaction traces.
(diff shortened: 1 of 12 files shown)


reentrancy-counter/README.md
```diff
@@ -0,0 +1,27 @@
+# High Gas Agent
+
```
rachit2501, line 2: Readme needs to be updated

reentrancy-counter/src/agent.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+        createTrace("0x3", [0, 0, 0]),
+      ]);
+      const findings: Finding[] = await handleTransaction(tx);
+      expect(findings).toStrictEqual([]);
+    });
+
+    it("Should ignore non reentrant calls", async () => {
+      const tx: TransactionEvent = createTxEvent([
+        createTrace("0x0", []),
+        createTrace("0x1", [0]),
```
rachit2501, line 57: More description on what 0, 1 ... represents ( array elements ) would be nice to have in comments

reentrancy-counter/src/agent.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+        createTrace("0x2", [0, 0]),
+        createTrace("0x3", [0, 0, 0]),
+        createTrace("0x2", [0, 0, 0, 0]),
+        createTrace("0x3", [0, 0, 0, 0, 0]),
+        createTrace("0x2", [0, 0, 0, 0, 0, 0]),
+        createTrace("0x4", [0, 1]),
+        createTrace("0x5", [0, 1, 0]),
+        createTrace("0x6", [0, 1, 0, 0]),
+        createTrace("0x4", [0, 1, 0, 0, 0]),
```
rachit2501, line 89: can you add a small diag showing the loops/flow, a link to some diag would be great.

rachit2501 (changes requested):

reentrancy-counter/src/agent.spec.ts
e1Ru1o: as you can see in the `createTrace` function the array represent the `traceAddress` field of a `Trace` that is why no explanation was added there, but in short the array represent the sequence of calls made that originated the `Trace` and the number represent the depth of each call.
You can read about it here https://openethereum.github.io/JSONRPC-trace-module

reentrancy-counter/src/agent.spec.ts
e1Ru1o: Check the link above an tell me if the diag is still needed.

rachit2501: Yes. I read the link. Would be better if you can provide a trace similar to in the example.

rachit2501: A
  CALLs B
    CALLs G
  CALLs C
    CALLs G

e1Ru1o: Well, I think that the calls all self explanatory (the stack is been showed), but I am going to add some indentation in the calls in order to simulate the approach.

e1Ru1o: The indentation looks very bad, then some comments added int the test (note that calls have the same shape of the stacks). I hope it looks clearer for you now.

rachit2501 (approved):

PR #9, opened 2021-09-13
Function call detector template

e1Ru1o:
Template for detecting function calls given a function signature.

Options:
- Detect only function calls `from` specific callers
- Detect only function calls `to` specific targets

### DIFF
--- a/general-agent-module/src/function.calls.ts
+++ b/general-agent-module/src/function.calls.ts
@@ -7,7 +7,7 @@ import {
 import { FindingGenerator } from "./utils";
 import Web3 from "web3";
 
-const web3 = new Web3();
+const abi = new Web3().eth.abi;
 
 interface AgentOptions{
   from?: string;
@@ -48,7 +48,7 @@ const createFilter = (options: AgentOptions | undefined): Filter => {
     }
 
     if (options.functionSignature !== undefined) {
-      const expectedSelector: string = web3.eth.abi.encodeFunctionSignature(options.functionSignature);
+      const expectedSelector: string = abi.encodeFunctionSignature(options.functionSignature);
       const functionSelector: string = traceInfo.input.slice(0, 10);
       if(expectedSelector !== functionSelector)
         return false;
--- a/general-agent-module/src/function.calls.ts
+++ b/general-agent-module/src/function.calls.ts
@@ -1,0 +1,72 @@
+import { 
+  Finding, 
+  HandleTransaction, 
+  TransactionEvent, 
+  Trace,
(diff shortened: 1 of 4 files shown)


general-agent-module/src/function.calls.ts
```diff
@@ -0,0 +1,70 @@
...
+} from "forta-agent";
+import { FindingGenerator } from "./utils";
+import Web3 from "web3";
+
+const abi = new Web3().eth.abi;
+
+interface AgentOptions{
+  from?: string;
+  to?: string;
+  functionSignature: string;
```
mauricio1802, line 15: `functionSignature` is a must on your agent, why not set it a regular argument?

general-agent-module/src/function.calls.ts
```diff
@@ -0,0 +1,70 @@
...
+  return (traceInfo) => {
+    if (options.from !== undefined && options.from !== traceInfo.from) {
+      return false;
+    }
+
+    if (options.to !== undefined && options.to !== traceInfo.to) {
+      return false;
+    }
+
+    if (options.functionSignature !== undefined) {
```
mauricio1802, line 48: Why do you need this `if`? unctionSignature will always be provided

general-agent-module/src/tests.utils.ts
```diff
@@ -95,4 +96,9 @@ export class TestTransactionEvent extends TransactionEvent {
     this.addresses[address] = true;
     return this;
   }
+
+  public addTrace(trace: Trace): TestTransactionEvent {
+    this.traces.push(trace);
+    return this;
+  }
```
mauricio1802, line 103: I think the idea with these methods is to make tests more readable and smooth to write. If I still need to create the whole Trace object and passing it as an argument that goal is missed I think. I would rather add parameters to the function and create the trace inside.

mauricio1802 (changes requested):

general-agent-module/src/function.calls.ts
```diff
@@ -0,0 +1,70 @@
...
+        return false;
+    }
+
+    return true;
+  };
+};
+
+export default function provideFunctionCallsDetectorAgent(
+  findingGenerator: FindingGenerator,
+  agentOptions?: AgentOptions
```
mauricio1802, line 61: `agentOptions` shouldn't be optional because the template needs the `functionSignature`. I would recommend moving the `functionSignature` from the options.

mauricio1802 (changes requested):
I left a few comments.

mauricio1802 (approved):

PR #15, opened 2021-09-16
YFI governance changes agent

e1Ru1o:
Detect txns that change YFI governance address

This PR should be a draft until the `general-agents-module` gets published, or an alternative installation is decided.


CeliktepeMurat:
Shouldn't that be inside Yearn Finance agent suit?

e1Ru1o:
> Shouldn't that be inside Yearn Finance agent suit?

The yearn finance module is only for template AFAIK, that is why this agent is not included in it

mauricio1802 (approved):

PR #20, opened 2021-09-21
Esm join agent

CeliktepeMurat:
This PR introduces ESM Join Event Agent that alerts when greater than 2 MKR sent.


Maker-Agents-Suit-Module/ESMAgent/src/ESM_Join_Event_Agent.ts
```diff
@@ -0,0 +1,52 @@
...
+  FindingSeverity,
+  FindingType,
+  Log,
+} from 'forta-agent';
+
+export const MAKER_ESM_JOIN_EVENT_SIGNATURE = 'Join(address,uint256)';
+export const MKR_DECIMALS = 18;
+
+const filterLog = (log: Log): boolean => {
+  const value = Number(BigInt(log.data)) / 10 ** MKR_DECIMALS;
```
e1Ru1o, line 18: Why this double cast?
If you cast to `BigInt` in order to work with a big number I think that when you cast to `Number` you lose the ability of handling the big number. 

Suggestions:
- If working with `Number` is enough (I think it is not) then remove the `BignInt` cast:
`Number(log.data) / 10 ** MKR_DECIMALS`
- Other option is using `BigInt` in both operands:
`BigInt(log.data) / BigInt(10 ** MKR_DECIMALS)`

Maybe is better to change the comparison and work with `MKR * DECIMALS` to avoid the division

e1Ru1o (changes requested):

e1Ru1o (approved):

mauricio1802 (approved):

PR #26, opened 2021-09-23
MakerDAO Governance Module

e1Ru1o:
- [x] Hat Changes Detector
- [x] Lift Events Detector

Blocking / Waiting for:
- Known addresses


Maker-Agents-Suit-Module/Governance-Module/src/lift.events.ts
```diff
@@ -0,0 +1,57 @@
...
+    metadata: {
+      address: unknown,
+    },
+  });
+
+export const provideLiftEventsListener = (
+  alertId: string, 
+  contractAddress: string,
+  isKnown: AddressVerifier = isAddressKnown,
+  topic: string = LIFT_EVENT,
```
mauricio1802, line 33: Why do you need to receive the topic in the arguments, this handler is for detecting only `Lift` events, what is the point of receiving the topic as an argument?

Maker-Agents-Suit-Module/Governance-Module/src/lift.events.ts
```diff
@@ -0,0 +1,57 @@
...
+  return async (txEvent: TransactionEvent) => {
+    const findings: Finding[] = [];
+
+    if(!txEvent.addresses[contract])
+      return findings;
+
+    for(const log of txEvent.logs) {
+      if((log.address === contract) && (log.topics[0] === topic)){
+        const topic1: string = decodeSingleParam('address', log.topics[1]).toLowerCase();
+        const topic2: string = decodeSingleParam('address', log.topics[2]).toLowerCase();
```
mauricio1802, line 47: Do you need `toLowerCase()` here? AFAIK all the addresses coming from forta-agent types are in lower case. If your `decodeSingleParam` doesn't change the case you don't need it.

mauricio1802 (approved):

Maker-Agents-Suit-Module/Governance-Module/src/lift.events.ts
e1Ru1o: True, but I made the agent a little bit more general, because the `note` modifier is used in other functions and maybe later they ask for those events. Maybe I can move this implementation to a separate file, and make a new function that wraps it and pass the `LIFT_EVENT`. Or I can just deletes the topic argument as you suggest. What do you think?

Maker-Agents-Suit-Module/Governance-Module/src/lift.events.ts
e1Ru1o: all the addresses coming from forta-agent types are in lower case, but the decoded address is not lower case, I mean when web3 decode the address it is not returning a lower case one, seems like it returns a checksum address.
Check this code for example:
```javascript
const addr = Web3.utils.leftPad("0xc", 40);
const encoded = web3.eth.abi.encodeParameter('address', addr);
const decoded = web3.eth.abi.decodeParameter('address', encoded);
console.log(decoded);
```
On the other hand if you use `0xd` you will get a lower case address.

CeliktepeMurat (approved):

rachit2501 (approved):

PR #28, opened 2021-09-24
Big queued price deviation

mauricio1802:
This PR adds a handler on the Oracle Security Agent for checking Oracles that enqueue prices with big deviations


Maker-Agents-Suit-Module/OSMAgent/src/big.queued.price.deviation.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+  it("should returns empty findings if the correct event wasn't logged", async () => {
+    handleTransaction = provideBigQueuedPriceDeviationHandler(
+      CONTRACT_ADDRESSES
+    );
+
+    const txEvent: TransactionEvent = new TestTransactionEvent()
+      .addTrace({
+        from: CONTRACT_ADDRESSES[0],
+        input: PEEK_FUNCTION_SELECTOR,
+        output: web3.eth.abi.encodeParameters(["uint128", "bool"], [108, true]),
```
e1Ru1o, line 164: Same here as above, I think that the problem is that you wrote tests for different `lenght` in the results of next and current, being consistent with the current implementation

Maker-Agents-Suit-Module/OSMAgent/src/big.queued.price.deviation.ts
```diff
@@ -0,0 +1,115 @@
...
+  contractAddress: string,
+  txEvent: TransactionEvent
+): Finding | null => {
+  const currentValues: bigint[] = getCurrentValues(contractAddress, txEvent);
+  const nextValues: bigint[] = getNextValuesForOSM(
+    contractAddress,
+    txEvent.traces
+  );
+
+  const lessLenght: number = Math.min(currentValues.length, nextValues.length);
```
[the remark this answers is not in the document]
mauricio1802: You are right, I am going to remove that line.

CeliktepeMurat (approved):

Maker-Agents-Suit-Module/OSMAgent/src/big.queued.price.deviation.ts
```diff
@@ -0,0 +1,115 @@
...
+  for (let i = 0; i < lessLenght; i++) {
```
CeliktepeMurat, line 95: May you elaborate this length topic. I'm just confused about lessLength and loop over it etc.

mauricio1802: I've already removed it. I thought I could have multiple lengths and in that case I wanted to iterate until one of the arrays were consumed. But both arrays must have the same length.

e1Ru1o (approved):

rachit2501 (approved):

PR #61, opened 2021-10-19
Vesper Agent 6: Check for updates in collateralFactorMantissa

avalonche:


Vesper-Agents-Suite/Compound-Collateral-Factor-Agent/src/index.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  beforeAll(() => {
+    handleTransaction = agent.handleTransaction
+    web3 = new Web3()
+  })
+
+  describe("handleTransaction", () => {
+    it("returns a finding if collateral update event has occured", async () => {
+      const data = web3.eth.abi.encodeParameters(
+        ['address', 'uint256', 'uint256'],
```
rachit2501, line 22: use abi methods exported from `forta-agent-tools` to avoid setting up web3 for encrytpion/decryption purposes.

Vesper-Agents-Suite/Compound-Collateral-Factor-Agent/src/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+// An event emitted when new collateralization factor is made for a cToken
+export const NEW_COLLATERAL_FACTOR_SIGNATURE =
+    'NewCollateralFactor(address,uint256,uint256)'
+
+export const COLLATERAL_FACTOR_EVENT_ALERT_ID = 'Vesper-6'
+
+export const generateHash = (signature: string): string => {
+    const hash = keccak256(signature).toString('hex');
+    return '0x' + hash;
+}
```
e1Ru1o, line 15: Seem like you are not using this function, then you can remove it
```suggestion
```

Vesper-Agents-Suite/Compound-Collateral-Factor-Agent/src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+import {
+  FindingGenerator,
+  provideEventCheckerHandler
+} from 'forta-agent-tools';
+import Web3 from 'web3';
+import { COLLATERAL_FACTOR_EVENT_ALERT_ID, COMPOUND_COMPTROLLER_ADDRESS, NEW_COLLATERAL_FACTOR_SIGNATURE } from './utils';
+
+const web3 = new Web3();
+
+const createFindingGenerator = (): FindingGenerator => {
```
e1Ru1o, line 15: I think that you can change this a little.

Right now you have a function that return another function, but the outer one doesn't receive any parameter and doesn't do any changes in the inner one. Then you can keep just the inner function.

The other option some parameters to the outer one. We usually pass the `alertId` as a parameter when we create a function like this, in your case it would be something like:
```typescript
const findingGenerator = createFindingGenerator(COLLATERAL_FACTOR_EVENT_ALERT_ID);
```

e1Ru1o, line 15: I think that you can change this a little.

Right now you have a function that return another function, but the outer one doesn't receive any parameter and doesn't do any changes in the inner one. Then you can keep just the inner function.

The other option some parameters to the outer one. We usually pass the `alertId` as a parameter when we create a function like this, in your case it would be something like:
```typescript
const findingGenerator = createFindingGenerator(COLLATERAL_FACTOR_EVENT_ALERT_ID);
```

Vesper-Agents-Suite/Compound-Collateral-Factor-Agent/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "forta-agent-starter",
+  "version": "0.0.1",
+  "description": "Forta Agent Typescript starter project",
```
e1Ru1o, line 4: From `forta-agent@0.019` on, `name` should be the `agentId`, then you can change it to `vesper-6`.
Also you can change the description.

Vesper-Agents-Suite/Compound-Collateral-Factor-Agent/src/agent.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+          NEW_COLLATERAL_FACTOR_SIGNATURE,
+          COMPOUND_COMPTROLLER_ADDRESS,
+          generateEvent(5),
+        )
+
+      const findings = await handleTransaction(txEvent)
+
+      expect(findings).toStrictEqual([
+        Finding.fromObject({
+          name: "COMPOUND NEW COLLATERAL FACTOR EVENT",
```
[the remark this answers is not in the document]
e1Ru1o: You are right. If you want to avoid writing the entire findings you can use your `createFindingGenerator` method.
We usually do this.

Vesper-Agents-Suite/Compound-Collateral-Factor-Agent/src/agent.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const handleTransaction: HandleTransaction = async (txEvent: TransactionEvent) => {
+  const agentHandler = provideEventCheckerHandler(
+    createFindingGenerator(),
+    NEW_COLLATERAL_FACTOR_SIGNATURE,
+    COMPOUND_COMPTROLLER_ADDRESS,
+  );
+
+  return agentHandler(txEvent);
+}
```
e1Ru1o, line 41: I think that you can check this function a bit in order to avoid the creation of the `agentHandler` each time that the agent receives a txn.

Options are:
- Add a provider:
```typescript
const provideHandleTransaction = (alertId) => provideEventCheckerHandler(...);
```
- Use the `provideEventCheckerHandler` to create the handler directly
```typescript
const handleTransaction: HandleTransaction = provideEventCheckerHandler(...)
```

Vesper-Agents-Suite/Compound-Collateral-Factor-Agent/src/agent.spec.ts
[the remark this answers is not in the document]
e1Ru1o: Yes, that is something that we discuss before in the team. The thing is that you are not testing the `createFindingGenerator`, you are testing the agent. The `createFindingGenerator` is always a pretty simple/elemental function and we assume that it is correct in order to write the tests a little bit cleaner or just to avoid writing the findings in their explicit form.

PR #72, opened 2021-11-02
Yearn Agent 1 - Multisigs

e1Ru1o:


Yearn-agents/Multisigs/src/test.utils.ts
```diff
@@ -0,0 +1,99 @@
+import { Finding, FindingSeverity, FindingType } from "forta-agent";
+
+const createOwnerAddedFindingGenerator = (ens: string, owner: string): Finding =>
```
avalonche, line 3: all findings have the same severity, type and protocol. I would make a generic finding generator and put the name, description, metadata etc. as parameters.

e1Ru1o: This can be done, but that severities & types are subject to changes and also in this case I prefer clarity/simplicity over the code refactor that you suggest. I am always in favor of the refactor, but this one is a minor one and I wouldn't like to apply that refactor in `utils` where the same happens.

Yearn-agents/Multisigs/src/test.utils.ts
```diff
@@ -0,0 +1,99 @@
...
+    alertId: "YEARN-1-5",
+    metadata: {
+      From: ens,
+      To: to,
+      Value: value.toString(),
+      TokenAddress: addr,
+    },
+  });
+
+const createETHTransferFindingGenerator = (ens: string, to: string, value: string): Finding =>
```
rachit2501, line 77: you can remove so many fn declarations, and use backtick (`) approach to accomplish this using a single function.

Yearn-agents/Multisigs/src/utils.ts
```diff
@@ -0,0 +1,152 @@
...
+      severity: FindingSeverity.Medium,
+      type: FindingType.Info,
+      protocol: "Yearn",
+      alertId: "YEARN-1-4",
+      metadata: {
+        Multisig: ens,
+        TxHash: txHash,
+        Payment: payment, 
+      },
+    });
```
rachit2501, line 75: too much unnecessary code rewriting.

rachit2501 (changes requested):

Yearn-agents/Multisigs/src/test.utils.ts
rachit2501: added some suggestion for refactor in my comment.

rachit2501 (approved):

mauricio1802 (approved):

CeliktepeMurat (approved):

PR #73, opened 2021-11-02
@yearn/agent8

rachit2501:


Yearn-agents/PPSValues/README.md
```diff
@@ -0,0 +1,30 @@
+# PPS Ambiguity Agent
+
+## Description
+
+This agent ambiguity in PPS ( Price per share )
```
avalonche, line 5:
```suggestion
This agent detects abnormality in PPS (Price per share)
```

Yearn-agents/PPSValues/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Describe each of the type of alerts fired by this agent
+
+- YEARN-8-1
+
+  - Fired when pps value decreases.
+  - Severity is always set to "high"
+  - Type is always set to "info"
+  - Metadata contains:
+  - `ppo`: Current Value of ppo,
```
avalonche, line 21: what's `ppo`?

Yearn-agents/PPSValues/README.md
```diff
@@ -0,0 +1,30 @@
...
+  - `tracker`: Older Value of ppo.
+
+- YEARN-8-2
+  - Fired when pps value change swiftly.
```
avalonche, line 25: could you define what is considered "swift change"

Yearn-agents/PPSValues/src/agent.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  it("returns empty findings if pps increases", async () => {
+    const blockEvent = new TestBlockEvent();
+
+    const findings = await handleBlock(blockEvent);
+
+    expect(findings).toStrictEqual([]);
+  });
+
+  it("Returns findings if pps decreases", async () => {
```
avalonche, line 26: could you make a test case for no change in pps

Yearn-agents/PPSValues/src/agent.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    const blockEvent = new TestBlockEvent();
+
+    const findings = await handleBlock(blockEvent);
+
+    expect(findings).toStrictEqual([
+      createFinding("1", "1.1", "Decrease in PPS", 1),
+    ]);
+  });
+
+  it("Returns findings if swift change in pps", async () => {
```
avalonche, line 36: test case for large change but has not reached the threshold

Yearn-agents/PPSValues/src/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+    ],
+  },
+] as AbiItem[];
+
+export const getYearnVaults = async (
+  web3: Web3,
+  blockNumber: string | number
+): Promise<string[]> => {
+  const yearnHelper = new web3.eth.Contract(helperAbi, yearnDataProvider);
+  return await yearnHelper.methods.assetsAddresses().call({}, blockNumber);
```
avalonche, line 24: is `await` statement needed?

Yearn-agents/PPSValues/src/agent.ts
```diff
@@ -0,0 +1,82 @@
...
+            pps.toString(),
+            tracker.toString(),
+            "Decrease in PPS",
+            1
+          )
+        );
+      }
+
+      // swift change in pps
+      if (Math.abs(pps.minus(tracker).dividedBy(tracker).toNumber()) > 0.1) {
```
avalonche, line 61: might be better to define the threshold 0.1 as a constant var at the beginning of the file so it is clear what the threshold is. If the threshold changes in the future, it'll be easy to locate and change.

Yearn-agents/PPSValues/src/mock.ts
```diff
@@ -0,0 +1,34 @@
+import { createAddress } from "forta-agent-tools";
+
+const mockPrice = jest.fn();
+mockPrice
+  .mockReturnValueOnce(1.1) // First call. Tracker = 101
+  .mockReturnValueOnce(1) // Second test. Tracker = 100  (<101)
+  .mockReturnValueOnce(100) // Third test. Tracker = 200. ( > 100 )
+  .mockReturnValueOnce(10); // Fourth test.
```
avalonche, line 8: is it possible to call this mock return value in the test case itself so it is clear what the price changed to in the test case?

Yearn-agents/PPSValues/src/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+};
+
+export const abi = [
```
avalonche, line 27: could you be a bit more clearer with the naming as to what this is the abi for? especially as there are two abis in the file called `helperAbi` and `abi`

Yearn-agents/PPSValues/src/agent.ts
```diff
@@ -0,0 +1,82 @@
...
+    type: FindingType.Unknown,
+    metadata: {
+      ppo,
+      tracker,
+    },
+  });
+};
+
+const provideHandleFunction = (web3: Web3): HandleBlock => {
+  let tracker = new BigNumber(1);
```
avalonche, line 35: why do we assume pps is 1 to start?

avalonche (commented):
it seems like your test cases are dependent on each other, making tests fail if a test case is added or out of order. This makes the test cases quite fragile and it is not obvious from the test file they are not independent.
I would make each test case independent by having a setup function and have an array of values the pps will move as the parameter so it's clear in each test case what the pps movement is.

Yearn-agents/PPSValues/src/utils.ts
rachit2501: yes. The call will return a promise.

Yearn-agents/PPSValues/src/agent.spec.ts
rachit2501: sure

Yearn-agents/PPSValues/src/agent.spec.ts
rachit2501: can you tell what you mean by threshold. If there is 10% increase or more, the finding will be triggered, else nothing.

Yearn-agents/PPSValues/src/agent.ts
rachit2501: Because we are calculating the increase in pps too. So if a value is 0, then any increment would result in infinity % change. I can add an exception condition, but didn't felt like it was required.

Yearn-agents/PPSValues/src/mock.ts
rachit2501: the jest.fn() is defined inside the class. Best i can do is export this and rewrite it in the tests.

rachit2501:
> it seems like your test cases are dependent on each other, making tests fail if a test case is added or out of order. This makes the test cases quite fragile and it is not obvious from the test file they are not independent. I would make each test case independent by having a setup function and have an array of values the pps will move as the parameter so it's clear in each test case what the pps movement is.

I think this is more on a preference side of things. I prefer to have a series of events to be triggered in a sequence in order to test the code, and making each individual event a test in itself. Also for the setup function and array approach, again, rather than creating new events in each "it" block, i will rather have dependent sequence of events, for me its easier to go with a top bottom flow, over creating a new independent flow every time.

Yearn-agents/PPSValues/src/agent.ts
```diff
@@ -0,0 +1,85 @@
...
+  tracker: string,
+  reason: string,
+  id: number
+) => {
+  return Finding.fromObject({
+    name: "Yearn PPS Agent",
+    description: `Year PPS value: ${reason}`,
+    alertId: `Yearn-8-${id}`,
+    severity: FindingSeverity.High,
+    type: FindingType.Unknown,
```
mauricio1802, line 26: You "Info" better than `unknow` or `Suspicious`

Yearn-agents/PPSValues/src/agent.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  return async (blockEvent: BlockEvent) => {
+    const findings: Finding[] = [];
+    const vaults = await getYearnVaults(web3, blockEvent.blockNumber);
+
+    for (let i = 0; i < vaults.length; i++) {
+      const vault = new web3.eth.Contract(vaultAbi as any, vaults[i]);
+
+      const pps = new BigNumber(
+        await vault.methods.getPricePerFullShare().call()
```
mauricio1802, line 46: You need to use the `blockNumber` for the call

mauricio1802, line 46: Don't await inside the the for, you can create all the promise and use `Promise.all` for await all of them, the performance should increase with that.

Yearn-agents/PPSValues/src/agent.ts
```diff
@@ -0,0 +1,85 @@
...
+          createFinding(
+            pps.toString(),
+            tracker.toString(),
+            "Very Swift change",
+            2
+          )
+        );
+      }
+
+      tracker = pps;
```
mauricio1802, line 75: I think you have to compare the `pps` of the same vault in different blocks. This looks like you are comparing the `pps` from multiple vaults between them.

Yearn-agents/PPSValues/src/utils.ts
mauricio1802: Actually, I think it is not needed, could you check?

mauricio1802 (changes requested):

mauricio1802 (changes requested):
Also, could you check the destination branch to `@yearn-agents`

Yearn-agents/PPSValues/src/agent.ts
rachit2501: nice observation. Thanks !

Yearn-agents/PPSValues/src/agent.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+      eth: {
+        Contract: MockWeb3.build_Mock(),
+      },
+    } as any;
+    handleBlock = agent.provideHandleFunction(mockWeb3 as any);
+  });
+
+  it("returns empty findings if pps increases", async () => {
+    const blockEvent = new TestBlockEvent();
+    mockPrice.mockReturnValueOnce(1.1);
```
e1Ru1o, line 20: I would like to suggest you to not use `1.1` or float numbers at all, because `pps` returns an `uint256`

rachit2501: Actually no. If you look at the code :
` return balance().mul(1e18).div(totalSupply());` 
Wad math is used, so any number will do.

Yearn-agents/PPSValues/src/agent.ts
```diff
@@ -0,0 +1,99 @@
...
+        findings.push(
+          createFinding(
+            pps.toString(),
+            vaultPrevValue.toString(),
+            "Very Swift change",
+            2
+          )
+        );
+      }
+      tracker[vaultAddress] = pps;
```
mauricio1802, line 89: You are storing the previous values every time you run the handler, this has a problem. Blocks are processed in an async way, so you don't have certainty about block N+1 will be processed after block N. I would encourage you to fetch the value for the previous block and use that for comparing. You should use cache for not fetching the same value multiple times

mauricio1802 (changes requested):

Yearn-agents/PPSValues/src/agent.spec.ts
e1Ru1o: Just to make my point clear, this is the `pps` function signature
```python
def pricePerShare() -> uint256:
```
you are mocking the value returned by that function

Yearn-agents/PPSValues/src/agent.ts
```diff
@@ -0,0 +1,140 @@
...
+
+const provideHandleFunction = (web3: Web3): HandleBlock => {
+  const threshold = 0.1;
+  const cache = new CacheContainer(new MemoryStorage());
+
+  return async (blockEvent: BlockEvent) => {
+    const findings: Finding[] = [];
+    let blockNumber = blockEvent.blockNumber;
+
+    const vaults = await getYearnVaults(web3, blockNumber);
```
CeliktepeMurat, line 48: I think you should use cache mechanism here to fetch vaults per block. You can take a look at the Mauricio's PR for example.

CeliktepeMurat (changes requested):
Left one comment

Yearn-agents/PPSValues/src/agent.ts
rachit2501: I can in id,e situation cache vaults too like the rest, but I was of impressions to check each block for new blocks too. Besides pops will change every block so I have to do the rest of the calls anyway,

mauricio1802 (approved):

Yearn-agents/PPSValues/src/agent.ts
CeliktepeMurat: I see, then LGTM!

CeliktepeMurat (approved):

PR #78, opened 2021-11-08
@yearn/agent#2

cuongquangnam:


CeliktepeMurat (changes requested):
Left some comments

Yearn-agents/debt-strategies/src/maker.fetcher.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    for (const strategy of strategies) {
+      const strategyContract = new this.web3.eth.Contract(
+        StrategyABI,
+        strategy
+      );
+
+      const name: string = await strategyContract.methods
+        .name()
+        .call({}, blockNumber);
```
mauricio1802, line 106: couldn't this be done with `promise.all`?

Yearn-agents/debt-strategies/src/agent.ts
```diff
@@ -0,0 +1,80 @@
...
+                type: FindingType.Info, 
+                alertId: "Yearn-2-1",
+                protocol: "Yearn", 
+                metadata: {
+                  strategy: maker.address
+                }
+              })
+            )
+          }
+          maker.checkedRatio = true
```
mauricio1802, line 58: Is this needed because you can have the same strategy in different vaults? If yes, wouldn't the findings related to calling `tend` be repeated? If no, why do you need this `checkedRatio`?

cuongquangnam: checked Ratio is used here to avoid repeating emitting the finding related to MakerVaultRatio for every tx in the same block.
The reason behind is that ratios are fetched based on the state of a block. If strategy.getCurrentMakerVaultRatio() < strategy.collateralizationRatio() - strategy.rebalanceTolerance() is true for a block, it will be true for every tx in the block.

CeliktepeMurat (approved):

mauricio1802 (approved):

PR #81, opened 2021-11-12
Yearn Agent 6 - Large Deposit/Withdraw

e1Ru1o:


Yearn-agents/Large-Deposit-Withdraw/src/agent.ts
```diff
@@ -0,0 +1,56 @@
...
+        vaults[i], 
+        depositLimit[i].multipliedBy(percent).div(100),
+      ));
+    }
+    for(let i = 0; i < vaults.length; ++i){
+      handlers.push(withdraw.provideLargeWithdrawDetector(
+        vaults[i], 
+        withdrawLimit[i].multipliedBy(percent).div(100),
+      ));
+    }
```
mauricio1802, line 42: Shouldn't you check the transfers from `want` token? Here you are checking the transfers from `vault` token, is that right?
Max withdrawn was computed with `totalAssets`, that refers to `want` token I think.

e1Ru1o: No, `totalAssets` is not being used, the value used is `totalSupply`.

mauricio1802 (approved):

CeliktepeMurat (approved):

Yearn-agents/Large-Deposit-Withdraw/src/agent.ts
```diff
@@ -0,0 +1,56 @@
...
+    );
+
+    const handlers: HandleTransaction[] = [];
+    for(let i = 0; i < vaults.length; ++i){
+      handlers.push(deposit.provideLargeDepositDetector(
+        vaults[i], 
+        depositLimit[i].multipliedBy(percent).div(100),
+      ));
+    }
+    for(let i = 0; i < vaults.length; ++i){
```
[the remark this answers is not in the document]
e1Ru1o: Sure I can, I just did two loops in order to group the two types of alerts instead of mixing them all. This doesn't affect the time complexity of then it should be fine.

rachit2501 (approved):

PR #82, opened 2021-11-17
@yearn/agent7

rachit2501:


Yearn-agents/InstantWithdrawl/README.md
```diff
@@ -0,0 +1,19 @@
+# Instant Withdraw Agent
+
+## Description
+
+To detect vaults that do not allow to withdraw a certain amount of funds and having a more precise idea of vault liquidity. Tris requires simulations where say we take holders that make up X% of vault assets and force them to withdraw their shares to see if the operation is doable in current conditions. DOes not need to be exact, but in the direction of "top 3 holders make 50% of the vault and can withdraw their funds" or "holder with 10% of vault assets cannot withdraw."
```
CeliktepeMurat, line 5: Small typo on 'Does'

Yearn-agents/InstantWithdrawl/src/agent.ts
```diff
@@ -0,0 +1,57 @@
+import { BlockEvent, Finding, HandleBlock, getJsonRpcUrl } from "forta-agent";
+import axios from "axios";
+import { generateFinding, getAccounts, Mapping } from "./utils";
+import Web3 from "web3";
+import runServer from "./blockchain";
+import { withdrawAbi } from "./abi";
+
+const providerHandleBlock = (Web3: any, axios: any): HandleBlock => {
```
CeliktepeMurat, line 8: I think you can get rid of 'any' types. E.g. (web3:Web3, _axios: axios)

CeliktepeMurat (changes requested):
Left 2 comments

Yearn-agents/InstantWithdrawl/src/agent.ts
rachit2501: There is no Axios type export from Axios. It's only for AxiosResponse and other similar.  Also for web3, that also wont work because, the Web3 variable I am passing in as the argument, is not an Object of Type Web3, but the class Web3 itself.
This is what vscode throws:
`Argument of type 'typeof Web3' is not assignable to parameter of type 'Web3'.
  Type 'typeof Web3' is missing the following properties from type 'Web3': defaultAccount, defaultBlock, currentProvider, setProvider, and 5 more.ts(2345)
agent.ts(55, 36): Did you mean to use 'new' with this expression?`

Yearn-agents/InstantWithdrawl/src/agent.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+
+      expect(findings).toStrictEqual([]);
+    });
+
+    it("Another account withdraws: If the withdraw is not working, return an alert", async () => {
+      const blockEvent = new TestBlockEvent();
+
+      const findings = await handleBlock(blockEvent);
+
+      expect(findings).toStrictEqual([
```
e1Ru1o, line 33: I know that maybe your mocks have all the information to let this test works as expected. I think this is the same issue that we discuss multiple times. The thing is that test should be self-descriptive, this test look exactly the same as the above one and is expecting a finding. Then as you can guess, finding how this test works requires not only look at the previous tests but also look at the mocks and maybe other things. Self-descriptive test are always better.

Yearn-agents/InstantWithdrawl/src/agent.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+        ),
+      ]);
+    });
+
+    it("Another account withdraws: If the withdraw is not working, return an alert", async () => {
+      const blockEvent = new TestBlockEvent();
+
+      const findings = await handleBlock(blockEvent);
+
+      expect(findings).toStrictEqual([
```
e1Ru1o, line 47: And the same happens here. Is like writing a tests suite that have all the tests written in other place.

Yearn-agents/InstantWithdrawl/src/agent.ts
```diff
@@ -0,0 +1,59 @@
...
+  return async (blockEvent: BlockEvent) => {
+    const findings: Finding[] = [];
+
+    const { mapping, users } = await getAccounts(axios);
+
+    const vaults = Object.keys(mapping);
+
+    const server = await runServer(Array.from(users) as string[]);
```
e1Ru1o, line 16: I am also wondering if this will work as expected. You are using the `port: 9545` and the agent can handle many block asynchronously, then if all the agents are trying to mount a server at the same port, I guess it will fail.

Yearn-agents/InstantWithdrawl/.env.example
```diff
@@ -0,0 +1 @@
+jsonRPC=
```
e1Ru1o, line 1: Is this needed?

Yearn-agents/InstantWithdrawl/README.md
```diff
@@ -0,0 +1,19 @@
+# Instant Withdraw Agent
+
+## Description
+
+To detect vaults that do not allow to withdraw a certain amount of funds and having a more precise idea of vault liquidity. Tris requires simulations where say we take holders that make up X% of vault assets and force them to withdraw their shares to see if the operation is doable in current conditions. Does not need to be exact, but in the direction of "top 3 holders make 50% of the vault and can withdraw their funds" or "holder with 10% of vault assets cannot withdraw."
```
e1Ru1o, line 5:
```suggestion
To detect vaults that do not allow to withdraw a certain amount of funds and having a more precise idea of vault liquidity. This requires simulations where we take holders that make up X% of vault assets and force them to withdraw their shares to see if the operation is doable in current conditions. Does not need to be exact, but in the direction of "top 3 holders make 50% of the vault and can withdraw their funds" or "holder with 10% of vault assets cannot withdraw."
```

Yearn-agents/InstantWithdrawl/src/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+  return Finding.fromObject({
+    name: "Yearn-agent-7",
+    description: `The ${index} account wasn't able to make the withdraw on vault ${vaultAddress}`,
+    severity: FindingSeverity.Info,
+    type: FindingType.Unknown,
+    alertId: "Yearn-agent-7",
+    metadata: {
+      balance,
+    },
+  });
```
e1Ru1o, line 72: Please use the `address` instead of the `index`. Also make the description simpler by moving `vaultAddress` and the `address` to the metadata

e1Ru1o (changes requested):
Is the expected behavior that the agent report lots of alerts in each block? Also in the README you mention some cases for the alerts, but the findings have none information about why are they being reported, i.e the reason of why that address is being analyzed, are the address in the top 3?, has the address 10% of the shares? <-- That kind of information.

Yearn-agents/InstantWithdrawl/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "forta-agent-starter",
+  "version": "0.0.1",
+  "description": "Forta Agent Typescript starter project",
```
e1Ru1o, line 4: Also update the `name` and `description`

e1Ru1o (changes requested):

Yearn-agents/InstantWithdrawl/.env.example
rachit2501: yes. To show the name of the env variable

Yearn-agents/InstantWithdrawl/src/agent.ts
rachit2501: good point. But that is not something that can be resolved by us. The best way for this would be for forta themselves provide a way to get a unique free port everytime. I already mentioned that in the feedback call.

Yearn-agents/InstantWithdrawl/src/agent.spec.ts
rachit2501: I can rewrite a more descriptive message sure. Otherwise the logic is correct.

Yearn-agents/InstantWithdrawl/src/utils.ts
rachit2501: I can move the vaultAddress to metadata for sure, but cant use address, the requirements need the number of top balance holders that can withdraw not the addresses itself. They want the index.

Yearn-agents/InstantWithdrawl/src/blockchain.ts
```diff
@@ -0,0 +1,32 @@
+import BigNumber from "bignumber.js";
+import path from "path";
+const ganache = require("ganache-cli");
+require("dotenv").config();
```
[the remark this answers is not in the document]
rachit2501: for a module to be imported in an es6 format, they need to be exported as one. Ganache isn't so needs to be imported as commonJs

Yearn-agents/InstantWithdrawl/src/agent.spec.ts
rachit2501: > Is the expected behavior that the agent report lots of alerts in each block? Also in the README you mention some cases for the alerts, but the findings have none information about why are they being reported, i.e the reason of why that address is being analyzed, are the address in the top 3?, has the address 10% of the shares? <-- That kind of information.

Because this cant be fixed. The graphql API returns result according to the current stage. If I import 150 address details, they would be acc to balances across all the vaults. So some vault may have very less and won't get listed. The only way is to push the number to like a thousand to decrease the probability, but that would result in a lot of unnecessary calls and stimulation.

Yearn-agents/InstantWithdrawl/src/agent.spec.ts
e1Ru1o: No, this time is not the same issue, is related with the old discussion, but in this case is the tests suite is not self-descriptive. The `agent.spec.ts` file itself doesn't have any information (in code) about what is being tested, it just have a label in each test.

Yearn-agents/InstantWithdrawl/src/utils.ts
e1Ru1o: Well, you can use both if the index is needed. Please add the `address` too.

Yearn-agents/InstantWithdrawl/src/blockchain.ts
```diff
@@ -0,0 +1,32 @@
...
+    networkId: 1,
+    port: 9545,
+    fork: process.env.jsonRPC,
+    hardfork: DEFAULT_HARDFORK,
+    fork_block_number: 12628504,
+    forkCachePath: FORK_TESTS_CACHE_PATH,
+    blockGasLimit: 210000,
+    locked: false,
+    genesisAccounts: [],
+    unlockedAccounts: accounts,
```
mauricio1802, line 27: I think the correct name is `unlocked_accounts`

Yearn-agents/InstantWithdrawl/src/agent.ts
mauricio1802: Actually, you can use `ganache.provider` instead of `ganache.server` for forking, in that way you don't need a node.

Yearn-agents/InstantWithdrawl/src/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+  return Finding.fromObject({
+    name: "Yearn-agent-7",
+    description: `The ${index} account wasn't able to make the withdraw on vault`,
+    severity: FindingSeverity.Info,
+    type: FindingType.Unknown,
+    alertId: "Yearn-agent-7",
+    metadata: {
+      balance,
+      vaultAddress,
+    },
```
mauricio1802, line 72: I think the finding should provide the percent that is available for withdrawing right now. I can't see that in the finding. Just reporting that index would require doing more computation after that, computations that we already done.

Yearn-agents/InstantWithdrawl/src/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+      id, account{
+        id
+      },
+      balancePosition,
+      balanceShares,
+      vault{
+        id
+      }
+    }
+  }`;
```
mauricio1802, line 25: Is there a way for specifying the block for this query?

Yearn-agents/InstantWithdrawl/src/blockchain.ts
```diff
@@ -0,0 +1,32 @@
+import path from "path";
+import dotenv from "dotenv";
+const ganache = require("ganache-cli");
```
mauricio1802, line 3: use `import ganache from "ganache-core"`

Yearn-agents/InstantWithdrawl/src/blockchain.ts
```diff
@@ -0,0 +1,32 @@
...
+const FORK_TESTS_CACHE_PATH = path.join(
+  __dirname,
+  "..",
+  "provider",
+  ".hardhat_node_test_cache"
+);
+
+export const DEFAULT_HARDFORK = "byzantium";
+
+export default function runServer(accounts: Array<string>) {
```
mauricio1802, line 15: I don't think using this `server` is the best approach. I think using `ganache.provider` is better.

mauricio1802 (changes requested):

Yearn-agents/InstantWithdrawl/src/utils.ts
rachit2501: No. Read the requirements first, please. We have to report the number of addresses that can withdraw, not the address. Its redundant information only, and is not needed at all.

Yearn-agents/InstantWithdrawl/hardhat.config.js
```diff
@@ -0,0 +1,14 @@
+dotenv = require("dotenv");
```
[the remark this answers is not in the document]
rachit2501: I tried both approaches. This file is not needed in the current version. Will remove it.

Yearn-agents/InstantWithdrawl/src/blockchain.ts
```diff
@@ -0,0 +1,32 @@
...
+  const options = {
+    chainId: 1,
+    networkId: 1,
+    port: 9545,
+    fork: process.env.jsonRPC,
+    hardfork: DEFAULT_HARDFORK,
+    fork_block_number: 12628504,
```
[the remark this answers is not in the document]
rachit2501: no. This means, fork the blockchain from this point onwards. Putting a limit to the block to transverse from just helps in search optimization that's all. Current block or not will not matter in this case. Please read about this in ganache docs

Yearn-agents/InstantWithdrawl/src/utils.ts
rachit2501: Yes. This was in my on my local machine already done. Let me push that.

Yearn-agents/InstantWithdrawl/src/blockchain.ts
rachit2501: The `ganache-core` is last updated 9 months ago. The latest updated one is `@ganache/core`. For that this is what I am getting:
image
So hence I used the `ganache-cli`

Yearn-agents/InstantWithdrawl/src/utils.ts
rachit2501: Yes.

Yearn-agents/InstantWithdrawl/src/agent.ts
rachit2501: i just checked this. Yes what you are saying can reduce the lines of code a bit. But if you see, if we don't spin a server ourselves, it is done internally. So sticking to this approach is better since it is much more clear in this way that a fork server is initiated.

rachit2501: Also, just using the provider wont make it work any way. Provider is an object we use in order to communicate to a particular rpc endpoint. So if you wanna stimulate, ie. fork, a server is mandatory. Using ganache.provider directly internally starts one that all.

Yearn-agents/InstantWithdrawl/src/agent.spec.ts
rachit2501: I dont get what you are trying to say.

Yearn-agents/InstantWithdrawl/src/blockchain.ts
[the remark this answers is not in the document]
mauricio1802: Not really. That means forking from that block, which means you will be adding transactions over that block. If you always use the same block, you'll be always using the same state for starting and you will be getting always the same results.

mauricio1802, line 22: This should be the block number where you are checking.

Yearn-agents/InstantWithdrawl/src/utils.ts
```diff
@@ -0,0 +1,78 @@
...
```
mauricio1802, line 25: I think you shouldn't get the first `150`, you should do the query per vault and get the top `5`, `10` or something like that.

Yearn-agents/InstantWithdrawl/src/agent.ts
```diff
@@ -0,0 +1,71 @@
...
+        promises.push(
+          contract.methods
+            .withdraw(balance, userAddress, 0)
+            .send({ from: userAddress })
+            .then(() => {
+              contract.methods
+                .balanceOf(userAddress)
+                .call({}, blockEvent.blockNumber)
+                .then((res: string) => {
+                  if (!(res === balance)) {
```
mauricio1802, line 46: why `!(x === y)` instead of `x !== y`?

mauricio1802, line 46: Are you sure about that condition? I think it should check if the balance after withdrawing (`res`) is bigger than zero or something like that

mauricio1802 (changes requested):

Yearn-agents/InstantWithdrawl/src/blockchain.ts
rachit2501: Not really.  https://ethereum.stackexchange.com/questions/90194/truffle-ganache-fork . The number specifies that the chain will fork from this blocknumber, but all the other new blocks are also included i.e. anything before this block is not accessbile, but everything after this is. And the web3 call is made to the latest block on the chain.

mauricio1802: Not really, forking from block `X` means that you will be adding blocks on top of block `X`. You can do a simple test for checking that.

e1Ru1o:
Follow up on this in #136

PR #89, opened 2021-12-07
DAO killing gauge

mauricio1802:


e1Ru1o (approved):

Curve-finance-suite/DAOKillingGauge/src/agent.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+const curveDAOAddress = createAddress("0x1");
+
+describe("Curve DAO Kill a Gauge Test Suite", () => {
+  let handleTransaction: HandleTransaction;
+
+  beforeAll(() => {
+    handleTransaction = providerHandleTransaction(curveDAOAddress);
+  });
+
+  it("should return empty findings method was not called", async () => {
```
yasmine9, line 15: i think this can be improved,
 'Should return empty findings when set_killed was not called' or
'Should return empty findings with empty transactions' for a more specific description.

Curve-finance-suite/DAOKillingGauge/src/agent.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+    const txEvent = new TestTransactionEvent();
+
+    const findings = await handleTransaction(txEvent);
+
+    expect(findings).toStrictEqual([]);
+  });
+
+  it("should return a finding method set_killed was called from DAO", async () => {
```
yasmine9, line 23: same previous note, we can use:
Should return a finding when set_killed method is called from DAO

Curve-finance-suite/DAOKillingGauge/src/agent.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+      from: curveDAOAddress,
+      input: gaugeInterface.encodeFunctionData("set_killed", [false]),
+    });
+
+    const findings = await handleTransaction(txEvent);
+
+    expect(findings).toStrictEqual([]);
+  });
+
+  it("should return empty findings if method set_killed is not called from DAO", async () => {
```
yasmine9, line 51: for a better precision we may use 
Should return empty findings when set_killed method is called from another address

PR #95, opened 2021-12-09
Unkill

mauricio1802:


Curve-finance-suite/Unkill/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+This agent detects when Curve Pool Owner calls `unkill_me` functions.
+
+## Supported Chains
+
+- Ethereum
+
+## Alerts
+
+Describe each of the type of alerts fired by this agent
```
Kalzak, line 13: I am unsure if this line a leftover from the README or something that should stay

Curve-finance-suite/Unkill/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+- CURVE-12
+  - Fired when Curve Pool Owner calls `unkill_me` function.
+  - Severity is always set to "info".
+  - Type is always set to "info".
+  - Metadata contains:
+    - `UnkillPool`: Contract where `unkill_me` was called.
```
Kalzak, line 20: If possible, a link to a TX on mainnet that would trigger this agent would be good to prove that it works not just with our test harness, but with real world data too.

Curve-finance-suite/Unkill/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "file": "npm run build && forta-agent run --file",
+    "publish": "forta-agent publish",
+    "push": "forta-agent push",
+    "disable": "forta-agent disable",
+    "enable": "forta-agent enable",
+    "keyfile": "forta-agent keyfile",
+    "test": "jest"
+  },
+  "dependencies": {
+    "bignumber.js": "^9.0.1",
```
RCantu92, line 22: `bignumber.js` could be uninstalled since it is not used, no?

Curve-finance-suite/Unkill/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "forta-agent": "^0.0.32",
+    "forta-agent-tools": "^1.0.31"
+  },
+  "devDependencies": {
+    "@types/jest": "^27.0.1",
+    "@types/nodemon": "^1.19.0",
```
RCantu92, line 28: `@types/nodemon` also not used.

Curve-finance-suite/Unkill/src/abi.ts
```diff
@@ -0,0 +1,5 @@
+import { utils } from "ethers";
```
RCantu92, line 1: `ethers` not installed in `package.json`.

Curve-finance-suite/Unkill/README.md
```diff
@@ -0,0 +1,20 @@
+# UnkillMe Agent
```
RCantu92, line 1: Could include a tx hash to test the agent with mainnet data.

Curve-finance-suite/Unkill/src/agent.spec.ts
```diff
@@ -0,0 +1,90 @@
+import {
+  FindingType,
+  FindingSeverity,
+  Finding,
```
RCantu92, line 4: `FindingType`, `FindingSeverity`, and `Finding` don't need to be imported since they are not being used.

Curve-finance-suite/Unkill/src/agent.ts
```diff
@@ -0,0 +1,37 @@
...
+  HandleTransaction,
+  FindingSeverity,
+  FindingType,
+} from "forta-agent";
+import { provideFunctionCallsDetectorHandler } from "forta-agent-tools";
+import { stablePoolInterface } from "./abi";
+
+const CURVE_POOL_OWNER = "0xeCb456EA5365865EbAb8a2661B0c503410e9B347";
+
+export const createFinding = (functionCallInfo: any): Finding => {
```
Kalzak, line 12: There is no need for `{}` brackets on the arrow function since there is an immediate return
Removing the `{}` brackets would then make the `return` keyword redundant because the return is implicit when there are no curly brackets on an arrow function

Curve-finance-suite/Unkill/src/agent.ts
```diff
@@ -0,0 +1,37 @@
...
+  poolOwnerAddress: string
+): HandleTransaction =>
+  provideFunctionCallsDetectorHandler(
+    createFinding,
+    stablePoolInterface.getFunction("unkill_me").format(),
+    { from: poolOwnerAddress }
+  );
+
+export default {
+  handleTransaction: provideHandleTransaction(CURVE_POOL_OWNER),
```
Kalzak, line 36: From my experience it was always standard to have the `alertId` as an input here so that the `alertId` could be changed by whoever is using the agent if necessary

Curve-finance-suite/Unkill/src/agent.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+const POOL_OWNER = createAddress("0x1");
+
+describe("Unkill Agent Test Suite", () => {
+  let handleTransaction: HandleTransaction;
+
+  beforeAll(() => {
+    handleTransaction = provideHandleTransaction(POOL_OWNER);
+  });
+
+  it("should return empty findings if nothing happens", async () => {
```
RCantu92, line 20: Maybe being more specific with what "nothing happens" means?

Change to something like `should return empty Findings because of non-relevant transactions`?

Curve-finance-suite/Unkill/src/agent.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+import { provideHandleTransaction, createFinding } from "./agent";
+import { stablePoolInterface } from "./abi";
+
+const POOL_OWNER = createAddress("0x1");
+
+describe("Unkill Agent Test Suite", () => {
+  let handleTransaction: HandleTransaction;
+
+  beforeAll(() => {
+    handleTransaction = provideHandleTransaction(POOL_OWNER);
```
Kalzak, line 17: Is is necessary for `handleTransaction` to be reset after every test?
From what I can see, there is no way that it would change between each test or affect the outcome of each test in any way.

Kalzak (commented):
I have commented with some suggestions and queries for this agent.

Curve-finance-suite/Unkill/src/agent.spec.ts
```diff
@@ -0,0 +1,90 @@
+import {
+  FindingType,
+  FindingSeverity,
+  Finding,
+  HandleTransaction,
+} from "forta-agent";
+import { createAddress, TestTransactionEvent } from "forta-agent-tools";
+import { provideHandleTransaction, createFinding } from "./agent";
```
RCantu92, line 8: Wow, importing `createFinding` rather than re-writing another instance in the test file. That's great.

Curve-finance-suite/Unkill/README.md
e1Ru1o: Should be removed :+1:

Curve-finance-suite/Unkill/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Description
+
+This agent detects when Curve Pool Owner calls `unkill_me` functions.
```
yasmine9, line 5: So this is detecting `unkill_me` calls that have been initiated through the Pool Owner contract? (which should be similar to detecting`unkill_me(_pool: address)` function calls on the Pool Owner contract )
Do we have cases where the function is called by the owner on the pool directly?  if they exist then they are not detected here

Curve-finance-suite/Unkill/src/agent.ts
e1Ru1o: Yes that is one old standard we use in agent templates, not needed when creating common agents template.

Curve-finance-suite/Unkill/src/agent.spec.ts
e1Ru1o: That code is executed only once before running the tests suite. It is a `beforeAll`, maybe you are thinking in a `beforeEach`.

Curve-finance-suite/Unkill/src/abi.ts
e1Ru1o: `ethers` is installed by default I think, no need to be added there

Curve-finance-suite/Unkill/README.md
mauricio1802: Sorry, I wasn't able to find any relevant transaction

Curve-finance-suite/Unkill/src/agent.ts
mauricio1802: I prefer to leave this decision to my formatter so my code will always have the same style, but I agree we should create a common formatter configuration and everyone should use it.

Curve-finance-suite/Unkill/src/agent.ts
mauricio1802: Actually using `alertId` as an argument was a introduced when creating `templates` as @e1Ru1o mentioned. For regular agents I think is cleaner to write them like this.

Curve-finance-suite/Unkill/package.json
mauricio1802: Actually, it is used in `start:dev` command

Curve-finance-suite/Unkill/README.md
mauricio1802: Sorry, I wasn't able to find on relevant transaction.

Curve-finance-suite/Unkill/README.md
mauricio1802: My understanding is the owner is the same for all the pools, which is `Pool Owner`, that is why I am monitoring it.

e1Ru1o (approved):

RCantu92 (approved):

PR #100, opened 2021-12-10
@curve/registry-remove-pool

Kalzak:
Agent being moved to its own directory and refactored to follow code format in this PR.

### DIFF
--- a/Curve-finance-suite/Registry-Remove-Pool/Dockerfile
+++ b/Curve-finance-suite/Registry-Remove-Pool/Dockerfile
@@ -1,0 +1,21 @@
+# Build stage: compile Typescript to Javascript
+FROM node:12-alpine AS builder
+WORKDIR /app
+COPY . .
+RUN npm ci
+RUN npm run build
+# obfuscate compiled Javascript (optional)
+# RUN npm install -g javascript-obfuscator
+# RUN javascript-obfuscator ./dist --output ./obfuscated --split-strings true --split-strings-chunk-length 3
+
+# Final stage: copy compiled Javascript from previous stage and install production dependencies
+FROM node:12-alpine
+ENV NODE_ENV=production
+WORKDIR /app
+# if using obfuscated code:
+# COPY --from=builder /app/obfuscated ./src
+# else if using unobfuscated code:
+COPY --from=builder /app/dist ./src
+COPY package*.json ./
+RUN npm ci --production
+CMD [ "npm", "run", "start:prod" ]
--- a/Curve-finance-suite/Registry-Remove-Pool/package.json
+++ b/Curve-finance-suite/Registry-Remove-Pool/package.json
@@ -1,0 +1,34 @@
+{
+  "name": "curve-13",
+  "version": "0.0.1",
+  "description": "Agent to detect a pool being removed from the Curve Registry",
+  "scripts": {
+    "build": "tsc",
+    "start": "npm run start:dev",
+    "start:dev": "nodemon --watch src --watch forta.config.json -e js,ts,json  --exec \"npm run build && forta-agent run\"",
+    "start:prod": "forta-agent run --prod",
+    "tx": "npm run build && forta-agent run --tx",
+    "block": "npm run build && forta-agent run --block",
+    "range": "npm run build && forta-agent run --range",
+    "file": "npm run build && forta-agent run --file",
+    "publish": "forta-agent publish",
+    "push": "forta-agent push",
+    "disable": "forta-agent disable",
+    "enable": "forta-agent enable",
+    "keyfile": "forta-agent keyfile",
+    "test": "jest"
+  },
+  "dependencies": {
+    "bignumber.js": "^9.0.1",
+    "forta-agent": "^0.0.32",
+    "forta-agent-tools": "^1.0.31"
+  },
+  "devDependencies": {
+    "@types/jest": "^27.0.1",
+    "@types/nodemon": "^1.19.0",
+    "jest": "^27.0.6",
+    "nodemon": "^2.0.8",
+    "ts-jest": "^27.0.3",
+    "typescript": "^4.3.4"
+  }
+}
--- a/Curve-finance-suite/Registry-Remove-Pool/package.json
+++ b/Curve-finance-suite/Registry-Remove-Pool/package.json
@@ -19,7 +19,6 @@
     "test": "jest"
   },
   "dependencies": {
-    "bignumber.js": "^9.0.1",
     "forta-agent": "^0.0.32",
     "forta-agent-tools": "^1.0.31"
   },
@@ -27,7 +26,6 @@
     "@types/jest": "^27.0.1",
     "@types/nodemon": "^1.19.0",
     "jest": "^27.0.6",
-    "nodemon": "^2.0.8",
     "ts-jest": "^27.0.3",
     "typescript": "^4.3.4"
   }
--- a/Curve-finance-suite/Registry-Remove-Pool/src/agent.spec.ts
+++ b/Curve-finance-suite/Registry-Remove-Pool/src/agent.spec.ts
@@ -1,0 +1,92 @@
+import {
+  Finding,
+  FindingSeverity,
+  FindingType,
+  HandleTransaction,
+  TransactionEvent,
+} from 'forta-agent';
+
+import { createAddress, TestTransactionEvent } from 'forta-agent-tools';
+
+import { provideHandleTransaction, REMOVE_POOL_SIGNATURE, R_IFACE } from './agent';
+
+const TARGET: string = createAddress('0xdead');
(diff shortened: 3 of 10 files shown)


Curve-finance-suite/Registry-Remove-Pool/src/agent.ts
```diff
@@ -0,0 +1,60 @@
...
+    Finding.fromObject({
+      name: 'Curve Registry contract called',
+      description: 'Event PoolRemoved has been emitted',
+      alertId: alertId,
+      severity: FindingSeverity.Info,
+      type: FindingType.Info,
+      protocol: 'Curve Finance',
+      metadata: {
+        // The address of the pool that was removed
+        pool_address: '0x' + metadata!.topics[1].substring(26,66),
```
yasmine9, line 38: We can use `decodeParameter` from forta-agent-tools here, 
```suggestion
        pool_address: decodeParameter('address',metadata!.topics[1]).toLowerCase(),
```

Curve-finance-suite/Registry-Remove-Pool/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+# RUN javascript-obfuscator ./dist --output ./obfuscated --split-strings true --split-strings-chunk-length 3
+
+# Final stage: copy compiled Javascript from previous stage and install production dependencies
+FROM node:12-alpine
+ENV NODE_ENV=production
+WORKDIR /app
+# if using obfuscated code:
+# COPY --from=builder /app/obfuscated ./src
+# else if using unobfuscated code:
+COPY --from=builder /app/dist ./src
```
yasmine9, line 18: You can remove unnecessary commented lines.

Curve-finance-suite/Registry-Remove-Pool/src/agent.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  TransactionEvent,
+} from 'forta-agent';
+
+import { createAddress, TestTransactionEvent } from 'forta-agent-tools';
+
+import { provideHandleTransaction, REMOVE_POOL_SIGNATURE, R_IFACE } from './agent';
+
+const TARGET: string = createAddress('0xdead');
+const ALERT_ID: string = 'registry-remove-pool-test';
+const RECEIVERS: string[] = [];
```
yasmine9, line 15: we can remove this const since it's never used

Curve-finance-suite/Registry-Remove-Pool/src/agent.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+        // This string is the data (not relevant for this test so it is empty)
+        '',
+        encodedPoolAddress,
+      );
+    const findings = await handler(tx);
+    expect(findings).toStrictEqual([
+      createFinding('0xab5801a7d398351b8be11c439e05c5b3259aec9b'),
+    ]);
+  });
+});
```
yasmine9, line 92: We can add a test case for  wrong event signature.

Curve-finance-suite/Registry-Remove-Pool/src/agent.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    expect(findings).toStrictEqual([]);
+  });
+
+  it('should ignore events from other contracts', async () => {
+    // Encode the pool address value
+    let encodedTopicData = R_IFACE.encodeFilterTopics(
+      R_IFACE.getEvent('PoolRemoved'),
+      ['0xab5801a7d398351b8be11c439e05c5b3259aec9b'],
+    );
+    const encodedPoolAddress: string = encodedTopicData[1] as string;
```
yasmine9, line 56: Maybe we can move this above to avoid repeating the same declaration on each test case.

Curve-finance-suite/Registry-Remove-Pool/src/agent.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+  it('should ignore empty transactions', async () => {
+    const tx: TransactionEvent = new TestTransactionEvent();
+
+    const findings = await handler(tx);
+    expect(findings).toStrictEqual([]);
+  });
+
+  it('should ignore events from other contracts', async () => {
```
RCantu92, line 50: Could be retitled to `should return no Findings from events in wrong contract`.

Curve-finance-suite/Registry-Remove-Pool/src/agent.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+        SENDERS[1],
+        // This string is the data (not relevant for this test so it is empty)
+        '',
+        encodedPoolAddress,
+    );
+    const findings = await handler(tx);
+    expect(findings).toStrictEqual([]);
+  });
+
+  it('should detect events from the target contract', async () => {
```
RCantu92, line 71: Could be retitled to `returns Findings from contract event emission`.

Curve-finance-suite/Registry-Remove-Pool/src/agent.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  const handler: HandleTransaction = provideHandleTransaction(
+    ALERT_ID,
+    TARGET,
+  );
+
+  beforeAll(() => {
+    for(let i = 0; i <= 30; ++i) {
+      RECEIVERS.push(createAddress(`0xa${i}`));
+      SENDERS.push(createAddress(`0xb${i}`));
+    }
```
RCantu92, line 40: Is creating so many addresses necessary? I only see one of these, `SENDERS[1]`, being used on L62.

Curve-finance-suite/Registry-Remove-Pool/src/agent.ts
```diff
@@ -0,0 +1,60 @@
...
+
+// Called for every transaction
+// Returns a list of findings (may be empty if no relevant events)
+export const provideHandleTransaction = (
+  alertId: string,
+  address: string,
+): HandleTransaction => {
+  return provideEventCheckerHandler(
+    createFindingGenerator(alertId),
+    REMOVE_POOL_SIGNATURE,
```
RCantu92, line 50: Could use `R_IFACE.getEvent('PoolRemoved').format('sighash')` since `R_IFACE` is already created.

Curve-finance-suite/Registry-Remove-Pool/src/agent.ts
```diff
@@ -0,0 +1,60 @@
...
+    address,
+  );
+}
+
+export default {
+  handleTransaction: provideHandleTransaction(
+    "curve-13",
```
RCantu92, line 57: We have been using the alert IDs in all caps, so this would be `CURVE-13`.

Curve-finance-suite/Registry-Remove-Pool/src/agent.ts
```diff
@@ -0,0 +1,60 @@
...
+]);
+
+// The address of the Curve registry from curve.readthedocs.io
+const REGISTRY_ADDRESS: string = '0x90E00ACe148ca3b23Ac1bC8C240C2a7Dd9c2d7f5';
+
+// Creates a finding generator to be used by provideEventCheckerHandler
+const createFindingGenerator = (alertId: string): FindingGenerator =>
+  (metadata: { [key: string]: any } | undefined): Finding => 
+    Finding.fromObject({
+      name: 'Curve Registry contract called',
```
RCantu92, line 30: Could change the Finding name to be more specific to the event being emitted rather than simply when the contract is called.

Something like `Pool Removed Event`.

Curve-finance-suite/Registry-Remove-Pool/src/agent.spec.ts
Kalzak: Having `SENDERS` and `RECEIVERS` is actually completely redundant now that I look at it. 
Both have been removed in favor of using `createAddress` for the occasional times where a different address is needed.

Curve-finance-suite/Registry-Remove-Pool/Dockerfile
Kalzak: If I had written these comments myself then definitely they should be removed, but since they were generated by Docker I feel inclined to leave them in as they do not affect the readability of the actual agent code itself since this is in its own isolated file.

Curve-finance-suite/Registry-Remove-Pool/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "curve-13",
```
e1Ru1o, line 2: `name` can't be shared between agents, so this agent should be mixed with `PoolAdded`

e1Ru1o (changes requested):

PR #114, opened 2022-01-06
control variable adjustment

mauricio1802:


OlympusDAO-Agent-Suit/BondContractControlVariableAdjustment/src/agent.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+        createAddress("0x12"),
+        encodeParameters(
+          ["uint256", "uint256", "uint256", "bool"],
+          [12, 11, 10, true]
+        )
+      );
+
+    const findings = await handleTransaction(txEvent);
+
+    expect(findings).toStrictEqual([]);
```
CeliktepeMurat, line 132: should not there be 1 finding here? On the txn, one of the event log is directed to the bond contract right?

CeliktepeMurat (commented):
Left just one comment

OlympusDAO-Agent-Suit/BondContractControlVariableAdjustment/src/agent.spec.ts
mauricio1802: No, because that contract wasn't found as a bond contract. The agent first will fetch the bonds length and after that will ask for the bond contract addresses, here I am saying that length is 2 and the contract emitting the event is the number 3 so it wasn't fetched as a bond contract.

CeliktepeMurat (approved):

e1Ru1o (approved):

PR #121, opened 2022-01-14
@pickle finance/idle funds

mauricio1802:


Pickle-Finance-Agents/IdleFunds/src/utils.ts
```diff
@@ -0,0 +1,100 @@
...
+  return pickleJar.available({ blockTag: blockNumber });
+};
+
+export const createFinding = (
+  pickleJar: string,
+  idleFundsPercent: string
+): Finding => {
+  return Finding.fromObject({
+    name: "Idle Funds",
+    description: "A pickle jar has a big amount of funds idle",
```
e1Ru1o, line 90:
```suggestion
    description: "A pickle jar has a big amount of idle funds",
```

e1Ru1o (commented):
I think you should parameterize the function `providerHandleBlock` a bit more by adding the threshold, because future changes in that threshold can make the test fail

e1Ru1o (approved):

PR #126, opened 2022-01-19
Range Management Monitor (RMM)

e1Ru1o:


Pickle-Finance-Agents/RangeManagementMonitor/src/mock.provider.ts
```diff
@@ -0,0 +1,42 @@
+import { Interface } from "@ethersproject/abi";
```
mauricio1802, line 1: I think you are not using this file

Pickle-Finance-Agents/RangeManagementMonitor/README.md
```diff
@@ -0,0 +1,22 @@
+# Pickle Range Management Monitor
+
+## Description
+
+Monitors the strategies to verify they are in range and earning fees.
+
+## Supported Chains
+
+- Ethereum
```
mauricio1802, line 9: Should we mention Ethereum? The keeper doesn't exist on ethereum.

e1Ru1o: Oh you mean that the current implementation doesn't work on mainnet, that is right. I added Ethereum because the agent will work on mainnet too if the params are changed accordingly (thinking that they want the agent in mainnet later).

I think it should be removed as you suggest, because in the other way we can list any compatible chain 🙃

Pickle-Finance-Agents/RangeManagementMonitor/src/mock.provider.ts
e1Ru1o: This one is used in the `addCallTo` function definition

e1Ru1o: Oh, I misread your comment, the file is used too, in `strategy.fetcher.spec.ts`

mauricio1802 (approved):

PR #132, opened 2022-01-25
Alpaca Guard

mauricio1802:


Alpaca-Finance-Agents/GuardAlert/src/agent.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+      .mockReturnValue(encodeParameter("bool", true));
+
+    const blockEvent = new TestBlockEvent();
+
+    const findings = await handleBlock(blockEvent);
+
+    expect(findings).toStrictEqual([]);
+  });
+
+  it("should return a finding if a price is deviated up", async () => {
```
RCantu92, line 84: Could be rewritten as `"should return a finding if price has deviated too high"`

Alpaca-Finance-Agents/GuardAlert/src/agent.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+      .calledWith(isCallToIsStable(workers[2]), undefined)
+      .mockReturnValue("0x08c379a0" + encodedError.slice(2));
+
+    const blockEvent = new TestBlockEvent();
+
+    const findings = await handleBlock(blockEvent);
+
+    expect(findings).toStrictEqual([
+      createFinding(lpTokens[1], PriceStatus.HIGH),
+    ]);
```
RCantu92, line 103: More for my own understanding, although you don't call any additional methods on `TestBlockEvent()`, because `ethersMock` has `callMock` as the value on one of its keys, that is how it is passed to `provideHandleBlock()`, right? Then that's where `findings` gets its values, right? Because `callMock` saves its state when called in L90 and L93. Am I understanding correctly?

Alpaca-Finance-Agents/GuardAlert/src/agent.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  it("should return a finding if a price is deviated down", async () => {
```
RCantu92, line 106: Could be rewritten as `"should return a finding if price has deviated too far down"`.

Alpaca-Finance-Agents/GuardAlert/src/agent.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+    const blockEvent = new TestBlockEvent();
+
+    const findings = await handleBlock(blockEvent);
+
+    expect(findings).toStrictEqual([
+      createFinding(lpTokens[0], PriceStatus.LOW),
+    ]);
+  });
+
+  it("should return multiple finindgs if multiple prices are deviated", async () => {
```
RCantu92, line 128: Misspelling on `findings` and whole sentence could be rewritten as `"should return multiple findings if multiple prices have deviated"`.

RCantu92 (commented):
No glaring mistakes IMO, just small grammar mistakes here and there. Learned a lot reading through it all. 👍

Alpaca-Finance-Agents/GuardAlert/src/utils.ts
```diff
@@ -0,0 +1,120 @@
...
+        return PriceStatus.LOW;
+
+      default:
+        return PriceStatus.UNKNOWN;
+    }
+  }
+};
+
+export const getLpsWorker = (vaults: Vault[]): string[][] => {
+  const workers: string[][] = [];
```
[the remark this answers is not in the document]
mauricio1802: It is better to use `Worker[]`, thanks for bringing this up

e1Ru1o (approved):

RCantu92 (approved):
Good job

PR #134, opened 2022-01-25
@alpaca/large position alert

RCantu92:
Agent that detects a large position being taken out of specific `Vault`s by listening to `Work` event emissions.

The `Vault`s and thresholds that signify they are a "large position" are as follows:
- More than 100,000 BUSD
- More than 30 ETH
- More than 3 BTCB
- More than 100,000 USDT
- More than 200,000 ALPACA
- More than 100,000 TUSD

### DIFF
--- a/Alpaca-Finance-Agents/Large-Position-Alert/README.md
+++ b/Alpaca-Finance-Agents/Large-Position-Alert/README.md
@@ -1,0 +1,27 @@
+# High Gas Agent
+
+## Description
+
+This agent detects transactions with high gas consumption
+
+## Supported Chains
+
+- Ethereum
+- List any other chains this agent can support e.g. BSC
+
+## Alerts
+
+Describe each of the type of alerts fired by this agent
+
+- FORTA-1
+  - Fired when a transaction consumes more gas than 1,000,000 gas
+  - Severity is always set to "medium" (mention any conditions where it could be something else)
+  - Type is always set to "suspicious" (mention any conditions where it could be something else)
+  - Mention any other type of metadata fields included with this alert
+
+## Test Data
+
+The agent behaviour can be verified with the following transactions:
+
+- 0x1b71dcc24657989f920d627c7768f545d70fcb861c9a05824f7f5d056968aeee (1,094,700 gas)
+- 0x8df0579bf65e859f87c45b485b8f1879c56bc818043c3a0d6870c410b5013266 (2,348,226 gas)
--- a/Alpaca-Finance-Agents/Large-Position-Alert/README.md
+++ b/Alpaca-Finance-Agents/Large-Position-Alert/README.md
@@ -29,4 +29,4 @@ This agent detects `Work` event emission from the `Vault` contract when the `loa
 
 The agent behaviour can be verified with the following transactions:
 
-- /*FIND EXISTING TXN THAT FITS CRITERIA*/
+- 0x7c14644d678090ed33d780fbf9fe016a26fea4d40af34ec226ffba1e554616ca (Closest `Work` event emission I could find. `loan` amount was ~26K BUSD)
--- a/Alpaca-Finance-Agents/Large-Position-Alert/README.md
+++ b/Alpaca-Finance-Agents/Large-Position-Alert/README.md
@@ -30,4 +30,4 @@ This agent detects `Work` event emission from the `Vault` contract when the `loa
 
 The agent behaviour can be verified with the following transactions:
 
-- [0xeda7ba552147a8a33f9f9b964c9d54605a1d20bcfc4d20e0c41cb6803f58246d](https://bscscan.com/tx/0xeda7ba552147a8a33f9f9b964c9d54605a1d20bcfc4d20e0c41cb6803f58246d) (`loan` amount was ~239K BUSD)
+- [0x811dc77a060dab7fe6fb06718f13a3abfc425a12f85867a2f2224c54e19a3266](https://www.bscscan.com/tx/0x811dc77a060dab7fe6fb06718f13a3abfc425a12f85867a2f2224c54e19a3266) (`loan` amount was ~631K BUSD)
--- a/Alpaca-Finance-Agents/Large-Position-Alert/README.md
+++ b/Alpaca-Finance-Agents/Large-Position-Alert/README.md
@@ -1,27 +1,32 @@
-# High Gas Agent
+# Large Position Agent
 
 ## Description
 
-This agent detects transactions with high gas consumption
+This agent detects `Work` event emission from the `Vault` contract when the `loan` meets any of the following criteria:
+  - More than 300 WBNB
+  - More than 100,00 BUSH
+  - More than 30 ETH
+  - More than 3 BTCB
+  - More than 100,000 USDT
+  - More than 200,000 ALPACA
+  - More than 100,000 TUSD
 
 ## Supported Chains
 
-- Ethereum
-- List any other chains this agent can support e.g. BSC
+- Binance Smart Chain (BSC)
 
 ## Alerts
 
-Describe each of the type of alerts fired by this agent
-
-- FORTA-1
-  - Fired when a transaction consumes more gas than 1,000,000 gas
-  - Severity is always set to "medium" (mention any conditions where it could be something else)
-  - Type is always set to "suspicious" (mention any conditions where it could be something else)
-  - Mention any other type of metadata fields included with this alert
+- ALPACA-1
+  - Fired when `Work` event is emitted with `loan` meeting any of the previously stated criteria.
+  - Severity is always set to "info."
+  - Type is always set to "unknown."
+  - Metadata contains:
+    - `id`: Id of the target position.
+    - `loan`: Amount to borrow from the vault.
 
 ## Test Data
 
 The agent behaviour can be verified with the following transactions:
 
-- 0x1b71dcc24657989f920d627c7768f545d70fcb861c9a05824f7f5d056968aeee (1,094,700 gas)
-- 0x8df0579bf65e859f87c45b485b8f1879c56bc818043c3a0d6870c410b5013266 (2,348,226 gas)
+- /*FIND EXISTING TXN THAT FITS CRITERIA*/
--- a/Alpaca-Finance-Agents/Large-Position-Alert/README.md
+++ b/Alpaca-Finance-Agents/Large-Position-Alert/README.md
@@ -29,4 +29,4 @@ This agent detects `Work` event emission from the `Vault` contract when the `loa
 
 The agent behaviour can be verified with the following transactions:
 
-- 0x7c14644d678090ed33d780fbf9fe016a26fea4d40af34ec226ffba1e554616ca (Closest `Work` event emission I could find. `loan` amount was ~26K BUSD)
+- [0xeda7ba552147a8a33f9f9b964c9d54605a1d20bcfc4d20e0c41cb6803f58246d](https://bscscan.com/tx/0xeda7ba552147a8a33f9f9b964c9d54605a1d20bcfc4d20e0c41cb6803f58246d) (`loan` amount was ~239K BUSD)
--- a/Alpaca-Finance-Agents/Large-Position-Alert/README.md
+++ b/Alpaca-Finance-Agents/Large-Position-Alert/README.md
@@ -19,11 +19,12 @@ This agent detects `Work` event emission from the `Vault` contract when the `loa
 
 - ALPACA-1
   - Fired when `Work` event is emitted with `loan` meeting any of the previously stated criteria.
-  - Severity is always set to "info."
-  - Type is always set to "unknown."
+  - Severity is always set to "Info."
+  - Type is always set to "Info."
   - Metadata contains:
     - `id`: Id of the target position.
     - `loan`: Amount to borrow from the vault.
+    - `vault`: Address of `Vault` the position is taken in.
 
 ## Test Data
 
--- a/Alpaca-Finance-Agents/Large-Position-Alert/README.md
+++ b/Alpaca-Finance-Agents/Large-Position-Alert/README.md
@@ -24,7 +24,7 @@ This agent detects `Work` event emission from the `Vault` contract when the `loa
   - Metadata contains:
     - `id`: Id of the target position.
     - `loan`: Amount to borrow from the vault.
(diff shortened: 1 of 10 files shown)


Alpaca-Finance-Agents/Large-Position-Alert/README.md
```diff
@@ -0,0 +1,32 @@
...
+## Supported Chains
+
+- Binance Smart Chain (BSC)
+
+## Alerts
+
+- ALPACA-1
+  - Fired when `Work` event is emitted with `loan` meeting any of the previously stated criteria.
+  - Severity is always set to "info."
+  - Type is always set to "unknown."
```
mauricio1802, line 23: Can we use type `Info`?

Alpaca-Finance-Agents/Large-Position-Alert/README.md
```diff
@@ -0,0 +1,32 @@
...
+  - Metadata contains:
```
mauricio1802, line 24: Can we are the `vault` where the position is being created?

Alpaca-Finance-Agents/Large-Position-Alert/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "forta-agent-starter",
+  "version": "0.0.1",
+  "description": "Forta Agent Typescript starter project",
```
mauricio1802, line 4: Update `name` and `description` please

Alpaca-Finance-Agents/Large-Position-Alert/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "publish": "forta-agent publish",
+    "push": "forta-agent push",
+    "disable": "forta-agent disable",
+    "enable": "forta-agent enable",
+    "keyfile": "forta-agent keyfile",
+    "test": "jest"
+  },
+  "dependencies": {
+    "bignumber.js": "^9.0.1",
+    "ethers": "^5.5.3",
```
mauricio1802, line 23: I think `ethers` is not needed. It is already installed with `forta-agent`

Alpaca-Finance-Agents/Large-Position-Alert/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "file": "npm run build && forta-agent run --file",
+    "publish": "forta-agent publish",
+    "push": "forta-agent push",
+    "disable": "forta-agent disable",
+    "enable": "forta-agent enable",
+    "keyfile": "forta-agent keyfile",
+    "test": "jest"
+  },
+  "dependencies": {
+    "bignumber.js": "^9.0.1",
```
mauricio1802, line 22: If we are not using this dependency, we should remove it.

Alpaca-Finance-Agents/Large-Position-Alert/src/agent.ts
```diff
@@ -0,0 +1,124 @@
+import BigNumber from 'bignumber.js'
```
mauricio1802, line 1: We can remove this import if it is not being used.

Alpaca-Finance-Agents/Large-Position-Alert/src/agent.ts
```diff
@@ -0,0 +1,124 @@
...
+  [VAULT_ADDRESSES[1], BigInt(30000000000000000000)],     // 30 ETH
+  [VAULT_ADDRESSES[2], BigInt(3000000000000000000)],      // 3 BTCB
+  [VAULT_ADDRESSES[3], BigInt(100000000000000000000000)], // 100,000 USDT
+  [VAULT_ADDRESSES[4], BigInt(200000000000000000000000)], // 200,000 ALPACA
+  [VAULT_ADDRESSES[5], BigInt(100000000000000000000000)]  // 100,000 TUSD
+])
+
+export const workEventAbi: string = 'event Work(uint256 indexed id, uint256 loan)';
+const VAULT_IFACE: utils.Interface = new utils.Interface([workEventAbi]);
+export const workEventSig: string = VAULT_IFACE.getEvent('Work').format('sighash');
```
mauricio1802, line 36: I think you could replace this 3 lines.
```suggestion
export const workEventSig: string = "Work(uint256,uint256)"
```

Alpaca-Finance-Agents/Large-Position-Alert/src/agent.ts
```diff
@@ -0,0 +1,124 @@
...
+  } else if(address === vaultAddresses[4].toLowerCase()) {
+    filteredAddress = vaultAddresses[4];
+    filteredThreshold = VAULT_MAP.get(vaultAddresses[4])
+  } else if(address === vaultAddresses[5].toLowerCase()) {
+    filteredAddress = vaultAddresses[5];
+    filteredThreshold = VAULT_MAP.get(vaultAddresses[5])
+  }
+
+  return [filteredAddress, filteredThreshold];
+}
```
mauricio1802, line 64: I don't get this function.

Alpaca-Finance-Agents/Large-Position-Alert/src/agent.ts
```diff
@@ -0,0 +1,124 @@
...
+      ["uint256", "uint256"],
+      metadata?.data
+    );
+
+    return Finding.fromObject({
+      name: "Large Position Event",
+      description: "Large Position Has Been Taken",
+      alertId: alertId,
+      severity: FindingSeverity.Info, // TODO: FIND OUT WHICH Severity TO USE HERE (Info?)
+      type: FindingType.Unknown,  // TODO: FIND OUT WHICH Type TO USE HERE (Unknown?)
```
mauricio1802, line 78: As mentioned in README, I think we can use `Info` here.

Alpaca-Finance-Agents/Large-Position-Alert/src/agent.ts
```diff
@@ -0,0 +1,124 @@
...
+      metadata:{
+        positionId: decodedData[0],
+        borrowAmount: decodedData[1]
```
mauricio1802, line 81: Adding the vault here could be useful.

Alpaca-Finance-Agents/Large-Position-Alert/src/agent.ts
```diff
@@ -0,0 +1,124 @@
...
+    });
+  };
+};
+
+export function provideHandleTransaction(
+  alertId: string,
+  addresses: string[],
+): HandleTransaction {
+  return async (txEvent: TransactionEvent): Promise<Finding[]> => {
+    let [filteredAddress, filteredThreshold] = filterForVault(addresses, txEvent.to);
```
mauricio1802, line 92: why not just `VAULT_ADDRESSES.get(txEvent.to)`? For knowing if it is not there I think you can just check if the result is `undefined`.

Alpaca-Finance-Agents/Large-Position-Alert/src/agent.ts
```diff
@@ -0,0 +1,124 @@
...
+          filteredEvent[0].data // TODO: CONFIRM THERE WILL ONLY BE ONE EVENT PER TXN, SINCE HARD CODING FIRST ITEM
+        );
+
+        if(filteredThreshold && decodedData[1] > filteredThreshold) {
+          return true;
+        } else {
+          return false;
+        }
+      }
+    );
```
mauricio1802, line 112: Here, you are missing multiple events, and you are also missing the event if the interaction was not directly with the vault emitting the event, because the only relevant address could be `txEvent.to`

mauricio1802 (changes requested):

Alpaca-Finance-Agents/Large-Position-Alert/src/agent.ts
```diff
@@ -0,0 +1,61 @@
...
+    const findings: Finding[] = [];
+    const workEvents = txEvent.filterLog(workEventAbi);
+
+    for(let i = 0; i < workEvents.length; i++) {
+      // NOTE: SOMEWHERE IN THE PROCESS, IT IS TURNING THE LETTERS IN THE VAULT ADDRESS LOWER CASE.
+      // SEE L11-L16 FOR IMPLEMENTATION.
+      const vaultThreshold = VAULT_MAP.get(workEvents[i].address);
+      if (vaultThreshold && BigInt(parseInt(workEvents[i].args["loan"])) > vaultThreshold) {
+        const createdFinding: Finding = createFinding(
+          parseInt(workEvents[0].args["id"]),
```
mauricio1802, line 47: Shouldn't this be 
```suggestion
          parseInt(workEvents[i].args["id"]),
```
?

Alpaca-Finance-Agents/Large-Position-Alert/src/agent.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+        name: "Large Position Event",
+        description: "Large Position Has Been Taken",
+        alertId: "ALPACA-1",
+        severity: FindingSeverity.Info,
+        type: FindingType.Info,
+        metadata: {
+          positionId: testPositionId.toString(),
+          loanAmount: testBorrowAmount.toString(),
+          vault: BUSD_VAULT_ADDRESS.toLowerCase()
+        }
```
mauricio1802, line 56: You could use your `createFinding` function here, but this way is ok too.

mauricio1802 (changes requested):
Try to keep consistency using single or double quotes.

Alpaca-Finance-Agents/Large-Position-Alert/src/agent.ts
e1Ru1o: I would like to also suggest to avoid using `provideEventChecker` form `forta-agent-tools` because this template uses `txEvent.filterEvent` and that function is being deprecated. Instead of that try `txEvent.filterLog`

Alpaca-Finance-Agents/Large-Position-Alert/src/agent.ts
```diff
@@ -0,0 +1,58 @@
...
+  return Finding.fromObject({
+    name: "Large Position Event",
+    description: "Large Position Has Been Taken",
+    alertId: "ALPACA-1",
+    severity: FindingSeverity.Info,
+    type: FindingType.Info,
+    metadata:{
+      positionId: posId.toString(),
+      loanAmount: loanAmount.toString(),
+      vault: vaultAddress?.toString()
```
e1Ru1o, line 30:
```suggestion
      vault: vaultAddress.toString()
```

Alpaca-Finance-Agents/Large-Position-Alert/src/agent.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export function provideHandleTransaction(vaultsMap: Map<string, bigint>): HandleTransaction {
+  return async (txEvent: TransactionEvent): Promise<Finding[]> => {
+    const findings: Finding[] = [];
+    const workEvents = txEvent.filterLog(workEventAbi);
+
+    for(let i = 0; i < workEvents.length; i++) {
+      const vaultThreshold = vaultsMap.get(workEvents[i].address);
+      if (vaultThreshold && BigInt(parseInt(workEvents[i].args["loan"])) > vaultThreshold) {
```
e1Ru1o, line 42: I would suggest to use `BigNumber` from `ethers` instead of `BigInt`. I am not sure about the correctness of using `BigInt(parseInt(value))` when value is a `ethers.Bignumberish`.

So I suggest to change to:
` vaultThreshold.lt(workEvents[i].args["loan"])` 
and change the mapping accordingly to `BigNumber` values.

Alpaca-Finance-Agents/Large-Position-Alert/src/agent.ts
```diff
@@ -0,0 +1,58 @@
...
+        const newFinding: Finding = createFinding(
+          parseInt(workEvents[i].args["id"]),
+          BigInt(parseInt(workEvents[i].args["loan"])),
```
e1Ru1o, line 45: same here, I like to work directly with the values returned by `ethers`

Alpaca-Finance-Agents/Large-Position-Alert/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "publish": "forta-agent publish",
+    "push": "forta-agent push",
+    "disable": "forta-agent disable",
+    "enable": "forta-agent enable",
+    "keyfile": "forta-agent keyfile",
+    "test": "jest"
+  },
+  "dependencies": {
+    "forta-agent": "^0.0.34",
+    "forta-agent-tools": "^1.0.32"
```
e1Ru1o, line 23: `forta-agent-tools` should be a dev-dependency

Alpaca-Finance-Agents/Large-Position-Alert/src/agent.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+    const txEvent: TransactionEvent = new TestTransactionEvent()
+      .addInvolvedAddresses(TEST_VAULT_ADDRESS, testMsgSender)
+      .addEventLog(badWorkSig, TEST_VAULT_ADDRESS, data);
+
+    const findings = await handleTransaction(txEvent);
+
+    expect(findings).toStrictEqual([]);
+  });
+
+  it("should return no Findings due to wrong contract ddress", async () => {
```
e1Ru1o, line 76:
```suggestion
  it("should return no Findings due to wrong contract address", async () => {
```

Alpaca-Finance-Agents/Large-Position-Alert/src/agent.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+    const wrongVaultAddress: string = createAddress("0x1111");
+
+    const txEvent: TransactionEvent = new TestTransactionEvent()
+      .addInvolvedAddresses(wrongVaultAddress, testMsgSender)
+      .addEventLog(workEventSig,wrongVaultAddress);
```
e1Ru1o, line 81: the idea in this test should be to use the correct event log and just a wrong address, so you should also add the `data`

e1Ru1o (changes requested):
Please add a test where multiple findings are detected. I would like to see tests with different `positionId`s, `loanAmount`s & `vault`s

mauricio1802 (approved):

PR #135, opened 2022-01-26
Check once every hour and set fund limit to 35%

virendrapatidar:
- Check condition once every hour (not every block)
- Increase limit to 35% from 20% for available fund.
- Have some duplicate code but that's due to old project structure.  All code in same directory act as one agent.

### DIFF
--- a/Vesper-Agents-Suite/Fund/src/agent.ts
+++ b/Vesper-Agents-Suite/Fund/src/agent.ts
@@ -3,12 +3,19 @@ import BigNumber from "bignumber.js";
 import abi from "./pool.abi";
 import Web3 from "web3";
 import { getTokensHere, getTotalValue, createFinding, getPools } from "./utils";
+import TimeTracker from "./time.tracker";
 
 const web3 = new Web3(getJsonRpcUrl());
+const tracker: TimeTracker = new TimeTracker();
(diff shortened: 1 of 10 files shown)


Vesper-Agents-Suite/Fund/src/agent.ts
```diff
@@ -28,8 +35,13 @@ function provideHandleFunction(web3: Web3): HandleBlock {
...
       const tokenHere = new BigNumber(value[2]);
-
-      if (tokenHere.isGreaterThan(totalValue.multipliedBy(0.2))) {
+      const [success, time] = tracker.tryGetLastTime(pool);
+      if (!success) {
+        // set this block as the time to start tracking the pool fund
+        tracker.update(pool, blockEvent.block.timestamp);
+      };
+      const elapsed: number = blockEvent.block.timestamp - time;
+      if (elapsed >= timeThreshold && tokenHere.isGreaterThan(totalValue.multipliedBy(0.35))) {
```
e1Ru1o, line 44: I think you should update the tracker inside this if:
```suggestion
      if (elapsed >= timeThreshold && tokenHere.isGreaterThan(totalValue.multipliedBy(0.35))) {
        tracker.update(pool, blockEvent.block.timestamp);
```

e1Ru1o (approved):

PR #138, opened 2022-01-28
@alpaca/liquidation alert

RCantu92:
`ALPACA-3` agent detects `Kill` event emissions from `Vault` contracts.
`ALPACA-4` agent detects `Kill` event emission from `Vault` contracts when `left` equals zero.

### DIFF
--- a/Alpaca-Finance-Agents/Liquidation-Alert/README.md
+++ b/Alpaca-Finance-Agents/Liquidation-Alert/README.md
@@ -1,0 +1,27 @@
+# High Gas Agent
+
+## Description
+
+This agent detects transactions with high gas consumption
+
+## Supported Chains
+
+- Ethereum
+- List any other chains this agent can support e.g. BSC
+
+## Alerts
+
+Describe each of the type of alerts fired by this agent
+
+- FORTA-1
+  - Fired when a transaction consumes more gas than 1,000,000 gas
+  - Severity is always set to "medium" (mention any conditions where it could be something else)
+  - Type is always set to "suspicious" (mention any conditions where it could be something else)
+  - Mention any other type of metadata fields included with this alert
+
+## Test Data
+
+The agent behaviour can be verified with the following transactions:
+
+- 0x1b71dcc24657989f920d627c7768f545d70fcb861c9a05824f7f5d056968aeee (1,094,700 gas)
+- 0x8df0579bf65e859f87c45b485b8f1879c56bc818043c3a0d6870c410b5013266 (2,348,226 gas)
--- a/Alpaca-Finance-Agents/Liquidation-Alert/README.md
+++ b/Alpaca-Finance-Agents/Liquidation-Alert/README.md
@@ -1,27 +1,31 @@
-# High Gas Agent
+# Liquidation Agent
 
 ## Description
 
-This agent detects transactions with high gas consumption
+This agent detects `Kill` event emissions from `Vault` contracts.
 
 ## Supported Chains
 
-- Ethereum
-- List any other chains this agent can support e.g. BSC
+- Binance Smart Chain (BSC)
 
 ## Alerts
 
-Describe each of the type of alerts fired by this agent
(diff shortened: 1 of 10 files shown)


Alpaca-Finance-Agents/Liquidation-Alert/src/agent.ts
```diff
@@ -0,0 +1,108 @@
...
+const createFindings = (
+  posId: number,
+  killer: string,
+  posOwner: string,
+  posVal: number,
+  debt: number,
+  prize: number,
+  left: number,
+  vaultAddress: string
+): Finding[] => {
```
mauricio1802, line 29: I think you shouldn't delegate the responsibility for deciding which alerts to emit to this function.

Alpaca-Finance-Agents/Liquidation-Alert/src/agent.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export function provideHandleTransaction(
+  addresses: string[]
+): HandleTransaction {
+  return async (txEvent: TransactionEvent): Promise<Finding[]> => {
+    const findings: Finding[] = [];
+    const killEvents = txEvent.filterLog(killEventAbi);
+
+    for(let i = 0; i < killEvents.length; i++) {
+      for(let v = 0; v < addresses.length; v++) {
```
mauricio1802, line 88: You have to remove this `for` loop. If you leave that for you would be checking if the address emitting the event is a relevant address `N` times, where `N` is the length of the `addresses` array. 
I think you can add one more address to the `TEST_VAULT_ADDRESSES` for checking what I am saying

mauricio1802 (changes requested):

mauricio1802 (approved):

Alpaca-Finance-Agents/Liquidation-Alert/src/agent.ts
```diff
@@ -0,0 +1,125 @@
...
+          killEvents[i].args["owner"],
+          killEvents[i].args["posVal"],
+          killEvents[i].args["debt"],
+          killEvents[i].args["prize"],
+          killEvents[i].args["left"],
+          killEvents[i].address
+        );
+        findings.push(newAgentThreeFinding);
+
+        if(Number(killEvents[i].args["left"]) === 0) {
```
e1Ru1o, line 101: This seems safer to me:
```suggestion
        if(BigNumber.from(0).eq(killEvents[i].args["left"])) {
```
I am not sure that parsing to `Number` will be safe.

You should import `BigNumber` from `ethers`:
`import { BigNumber } from "ethers";`

e1Ru1o (changes requested):

Alpaca-Finance-Agents/Liquidation-Alert/README.md
```diff
@@ -0,0 +1,46 @@
...
+  - Severity is always set to "Info."
+  - Type is always set to "Info."
+  - Metadata contains:
+    - `positionId`: Id of the target position.
+    - `positionkiller`: Address that executed the liquidation.
+    - `positionOwner`: Address of the owner of the target position.
+    - `positionValue`: Value amount of target position.
+    - `debt`: Debt value of the target position.
+    - `prize`: Combined amount of what is awarded to the liquidator and the treasury fees.
+    - `left`: What is returned to the target position's owner after liquidation execution.
```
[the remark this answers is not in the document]
RCantu92: That is true.

However, the other agent would also include it and perhaps including it in the one that confirms it is zero can display it for "peace of mind" for whoever sees it. Thoughts? I'm kind of leaning towards leaving it, especially if it doesn't hurt to include it.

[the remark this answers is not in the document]
e1Ru1o: Keep it

e1Ru1o (approved):

PR #141, opened 2022-02-02
AllocPoint Change Alert

RCantu92:
This agent detects when PancakeSwap's `Timelock` contract has a `QueueTransaction` event emitted from successful call to a pool's `set` function to update `allocPoint`. It also detects when a Mdex pool has its `set` function called by `BSCPool` to update `allocPoint`.


Alpaca-Finance-Agents/AllocPoint-Change-Alert/src/agent.ts
```diff
@@ -0,0 +1,113 @@
...
+  pools: Map<string, Map<number, string>>,
+  poolControllers: string[]
+): HandleTransaction {
+  return async (txEvent: TransactionEvent): Promise<Finding[]> => {
+    const findings: Finding[] = [];
+    const queueTxnEvents = txEvent.filterLog(pcsQueueTxnAbi);
+    // TODO: FIND DEPLOYED BoardRoomMDX TO FILTER BY ITS ADDRESS 
+    const mdexSetCalls = txEvent.filterFunction(mdexSetAbi);
+
+    if(queueTxnEvents.length > 0) {
```
mauricio1802, line 55: you can actually remove this `if`. When the length is `0` the `for` won't do any iteration.

Alpaca-Finance-Agents/AllocPoint-Change-Alert/src/agent.ts
```diff
@@ -0,0 +1,113 @@
...
+                  decodedData[2],
+                  queueTxnEvents[event].args["target"]
+                );
+  
+                findings.push(pcsFinding);
+              }
+            }
+          }
+        }
+      }
```
mauricio1802, line 81: Why not just check that a transaction was queued for calling `set(..)` on the masterchef contract?

Alpaca-Finance-Agents/AllocPoint-Change-Alert/src/agent.ts
```diff
@@ -0,0 +1,113 @@
...
+                mdexSetCalls[call].args["_withUpdate"],
+                txEvent.to
+              );
+              findings.push(mdexFinding);
+            }
+          }
+        }
+      }
+    }
+
```
mauricio1802, line 103: Why not just check if `set(....)` was called in the bscPool contract?

Alpaca-Finance-Agents/AllocPoint-Change-Alert/src/agent.ts
RCantu92: I'm not sure I'm understanding.

So instead of `txEvent.filterLog(pcsQueueTxnAbi)`, which checks for the `QueueTransaction` event emission from the `Timelock` contract, I could filter based on the info from the masterchef contract directly?

I think I  may not have fully understood the PancakeSwap architecture entirely.

Alpaca-Finance-Agents/AllocPoint-Change-Alert/src/agent.ts
```diff
@@ -0,0 +1,102 @@
...
+const queueTxnAbi: string = "event QueueTransaction(bytes32 indexed txHash, address indexed target, uint value, string signature, bytes data, uint eta)";
+
+
+function isAddressRelevant(contractAddress: string | null, poolArray: string[]): boolean {
+  if(contractAddress && poolArray.includes(contractAddress)) {
+    return true;
+  } else {
+    return false;
+  }
```
mauricio1802, line 29:
```suggestion
  return contractAddress && poolArray.includes(contractAddress) 
  }
```
When you see the patter `if (<boolean_expresion>) return true; else return false` that can be usually changed by `return <boolean_expresion>`

Alpaca-Finance-Agents/AllocPoint-Change-Alert/src/agent.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+function containsFuncSig(log: LogDescription, functionSig: string): boolean {
+  if(log.args["signature"] === functionSig) {
+    return true;
+  } else {
+    return false;
+  }
+}
```
mauricio1802, line 38:
```suggestion
function containsFuncSig(log: LogDescription, functionSig: string): boolean {
 return log.args["signature"] === functionSig);
}
```

Alpaca-Finance-Agents/AllocPoint-Change-Alert/src/agent.ts
```diff
@@ -0,0 +1,102 @@
...
+            log.args["data"]
+          );
+          return createFinding(
+            decodedData[0],
+            decodedData[1],
+            decodedData[2],
+            log.args["target"]
+          );
+        }),
+      ...txEvent.filterFunction(setFuncAbi)
```
mauricio1802, line 83: Shouldn't you use the second parameter for indicating where the function should be called?

mauricio1802 (changes requested):

mauricio1802 (approved):

Alpaca-Finance-Agents/AllocPoint-Change-Alert/src/agent.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export function provideHandleTransaction(
+  poolOwners: string[]
+): HandleTransaction {
+  return async (txEvent: TransactionEvent): Promise<Finding[]> => {
+    const findings: Finding[] = [];
+
+    findings.push(
+      ...txEvent.filterLog(queueTxnAbi)
+        .filter(log => isAddressRelevant(log.address, poolOwners))
```
e1Ru1o, line 60: I would suggest to always use lowers everywhere to avoid possible errors if the `POOL_OWNERS` array change. Right now I guess everything works well because the array is using checksumed addresses, but if the client change that list and is not aware of this the agent will not work as expected.

So maybe is safer to add a `.map(addr => addr.toLowerCase())` in the end of L16, and also use `log.addres..toLowerCase()` in the highlighted line above

e1Ru1o (commented):
Would like to also see a test case detecting multiple findings

e1Ru1o (changes requested):

Alpaca-Finance-Agents/AllocPoint-Change-Alert/src/agent.ts
```diff
@@ -0,0 +1,91 @@
...
+export function provideHandleTransaction(
+  ownerOne: string,
+  ownerTwo: string
+): HandleTransaction {
+  return async (txEvent: TransactionEvent): Promise<Finding[]> => {
+    const findings: Finding[] = [];
+
+    findings.push(
+      ...txEvent.filterLog(queueTxnAbi)
+        .filter(log => log.address.toLowerCase() === ownerOne)
```
e1Ru1o, line 55:
```suggestion
      ...txEvent.filterLog(queueTxnAbi, ownerOne)
```

e1Ru1o (approved):

PR #111, opened 2022-01-05
Olympus dao/bond ownershipchanges

CeliktepeMurat:
This PR introduces agent for ownership transfers on Bond Contracts.

PR #145, opened 2022-02-05
@impossible finance/flash loan

Kalzak:
# Flash Loans detector

## Types of changes

- [x] New Agent
- [ ] Fix an agent error

## New Agent PR readiness

- [ ] The `name` and `description` fields of `package.json` describe the agent properly
- [ ] The agent README is updated
- [ ] The agent README sufficiently describes what the agent does
- [ ] All the alerts that the agent emits are listed in the README
- [ ] Code is easy to follow and/or properly comment
- [ ] Agent contains tests
- [ ] All tests pass when running `npm tests`
- [ ] The code do what the README describes
- [ ] Findings `protocol` is specified (if the agent is for an specific one)
- [ ] the agent follows Forta recommended [best practices](https://docs.forta.network/en/latest/best-practices/)


Impossible-Agent-Suite/flash-loan/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "publish": "forta-agent publish",
+    "push": "forta-agent push",
+    "disable": "forta-agent disable",
+    "enable": "forta-agent enable",
+    "keyfile": "forta-agent keyfile",
+    "test": "jest"
+  },
+  "dependencies": {
+    "forta-agent": "^0.0.35",
+    "forta-agent-tools": "^1.0.32"
```
e1Ru1o, line 23: this should be a dev-dependency

Impossible-Agent-Suite/flash-loan/src/agent.ts
```diff
@@ -0,0 +1,88 @@
...
+  'function token0() external view returns (address)',
+  'function token1() external view returns (address)',
+];
+
+// Receives the address of a contract with same ABI as PAIR_SWAP_ABI and returns the factory address
+export const getContractFactory = async (swapContract: string) => {
+  const provider = getEthersProvider();
+  const contractInterface = new ethers.Contract(swapContract, PAIR_SWAP_ABI, provider);
+  return await contractInterface.factory();
+}
```
e1Ru1o, line 32: I think that you should add a `try {} catch {}` because you can found an address with the swap event & function but without the factory function (i.e at this point you are not sure of interacting with an impossible address)

Also adding a cache here will improve the performance of the agent. So please add a cache to this function.

e1Ru1o (changes requested):
Overall very good. 
I would also like to see a test detecting multiple findings in the same txn

Impossible-Agent-Suite/flash-loan/package.json
Kalzak: You're selected four lines here, but by looking at the outcome of a similar recommended change [here](https://github.com/NethermindEth/Forta-Agents/pull/111/files#diff-bfbd5e3d2ac2469ee61828991c4a035604b91dc7e3e4f7e05a24e637325ec9a3) I have only moved `forta-agent-tools` to `devDependencies`

Impossible-Agent-Suite/flash-loan/src/agent.ts
Kalzak: Both the `try {} catch {}` and cache have been implemented.

e1Ru1o (approved):

Impossible-Agent-Suite/flash-loan/src/agent.ts
```diff
@@ -0,0 +1,112 @@
...
+      // Query the smart contract for its factory
+      const provider = getEthersProvider();
+      const contractInterface = new ethers.Contract(swapContract, PAIR_SWAP_ABI, provider);
+      factory = await contractInterface.factory();
+      // Add it to the cache
+      cache.set(swapContract, factory);
+    } catch(error) {
+      // If the query fails (contract doesn't have a `factory` function then set to zero address
+      cache.set(swapContract, '0x0000000000000000000000000000000000000000');
+    }
```
e1Ru1o, line 50: I think you should set `factory` variable here too, just to not return undefined

Kalzak: Addressed in following commit.

Impossible-Agent-Suite/flash-loan/src/agent.ts
```diff
@@ -0,0 +1,129 @@
...
+      // If the query fails
+      cache.set(swapContract, false);
+      isFromFactory = false;
+    }
+  }
+  return isFromFactory;
+}
+
+export const provideHandleTransaction = (
+  address: string,
```
e1Ru1o, line 67: this variable is not being used. Please change its name, and pass it to the `checkFromFactory` function call in Line 98 in order to avoid using the global variable `SWAP_FACTORY_ADDRESS` in Line 47

Impossible-Agent-Suite/flash-loan/src/agent.ts
```diff
@@ -0,0 +1,129 @@
...
+                description: 'A flash loan has been executed on an Impossible Finance StableXSwap contract',
+                alertId: 'IMPOSSIBLE-5',
+                severity: FindingSeverity.Info,
+                type: FindingType.Info,
+                protocol: 'Impossible Finance',
+                metadata: {
+                  amount0Out: swapCall.args.amount0Out.toString(),
+                  amount1Out: swapCall.args.amount1Out.toString(),
+                  to: swapCall.args.to.toLowerCase(),
+                  data: ethers.BigNumber.from(swapCall.args.data).toHexString(),
```
e1Ru1o, line 113: I think `swapCall.args.data` is an hex string, isn't it?

yasmine9 (changes requested):

Impossible-Agent-Suite/flash-loan/src/agent.ts
Kalzak: That's right I've just turned it into a `BigNumber` and then back into `hex`. Will be addressed in the following commit.

Impossible-Agent-Suite/flash-loan/src/agent.ts
Kalzak: This will be addressed in the following commit.

Impossible-Agent-Suite/flash-loan/README.md
```diff
@@ -0,0 +1,32 @@
...
+This agent detects if a flash loan has been taken
+This agent detects flashloans through the function `swap` on pairs generated through `Swap Factory V1`
+
+## Supported Chains
+
+- Binance Smart Chain
+
+## Alerts
+
+Describe each of the type of alerts fired by this agent
```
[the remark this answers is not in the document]
Kalzak: Will be addressed in the following commit

Impossible-Agent-Suite/flash-loan/src/agent.ts
```diff
@@ -0,0 +1,129 @@
...
+} from 'forta-agent';
+
+const { ethers } = require('ethers');
+
+import LRU from 'lru-cache';
+
+const cache: LRU<string, boolean> = new LRU<string, boolean>({ max: 10000 });
+
+//let SWAP_FACTORY_ADDRESS = '0x918d7e714243f7d9d463c37e106235dcde294ffc';
+let SWAP_FACTORY_ADDRESS = '0x5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6f';
```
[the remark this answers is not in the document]
Kalzak: The line has been removed so only the Impossible Finance factory is in `agent.ts`. Users can manually add the other address after reading the `README` if they wish.

yasmine9 (approved):

PR #148, opened 2022-02-12
Adding gnosis Agent

0xrishabh:


gnosisAgent/src/function.monitoring.ts
```diff
@@ -0,0 +1,76 @@
...
+      let argumentsList:any[] = functionInvocation.functionFragment.inputs.map((obj)=>obj.name)
+
+      let metadata:any = {
+        by: txEvent.from,
+      }
+      for (let key in functionInvocation.args) {
+        if (argumentsList.includes(key)){
+          metadata[key] = functionInvocation.args[key].toString()
+        }
+      }
```
e1Ru1o, line 54: I think that this should work too:
```suggestion
      for (let key of argumentsList) {
        metadata[key] = functionInvocation.args[key].toString();
      }
```

gnosisAgent/README.md
```diff
@@ -0,0 +1,32 @@
+## Gnosis Safe admin changes
```
e1Ru1o, line 1: Please use the following sections in the Readme that are the standard template
- Title (you already have this one)
- Description (equivalent to Overview)
- Supported Chains
- Alerts
- Tests Data

e1Ru1o (changes requested):
Let me also remark that the agent is for Zodiac/Gnosis SafeSnap module. So please mention that in all the descriptions instead of only Gnosis.

e1Ru1o (approved):

PR #150, opened 2022-02-16
Cooldown Monitor Agent

RCantu92:
This agent detects when a question, based on its `question_id`, begins its cooldown period.


SafeSnap-Agents/Cooldown-Alert/src/agent.ts
```diff
@@ -0,0 +1,95 @@
...
+            questionIds[id],
+            finalizeTS,
+            blockEvent.blockNumber
+          )
+        );
+      }
+
+      // NOTE: IF ONLY CHECKING WHEN COOLDOWN BEGINS,
+      // NO NEED TO WAIT FOR QUESTION EXPIRATION FOR DELETION
+      delete questionIds[id];
```
e1Ru1o, line 81: This is not removing the questionId from the array, this is just changing the value in that position of the array to undefined. Also I think this must be done inside the if above, because you are always removing the questionId, so there is only one chance to check each question.

SafeSnap-Agents/Cooldown-Alert/src/agent.ts
```diff
@@ -0,0 +1,95 @@
...
+  FindingType,
+  getEthersProvider,
+  TransactionEvent,
+} from "forta-agent";
+import {
+  propQuestionCreateAbi,
+} from "./abi";
+import DataFetcher from "./data.fetcher";
+
+const REALITIO_ERC20: string = "0x8f1CC53bf34932591177CDA24723486205CA7510".toLowerCase();
```
e1Ru1o, line 16: This address you are using is not the oracle of the module you have, you need to fetch the oracle from the Module contract, because the only address we can have in the agent is the address of the module

SafeSnap-Agents/Cooldown-Alert/src/data.fetcher.ts
```diff
@@ -0,0 +1,30 @@
...
+    readonly realitio: string;
+    private rContract: Contract;
+    readonly provider: providers.Provider;
+    private cache: LRU<string, ResultType>;
+
+    constructor(realitio: string, provider: providers.Provider) {
+        this.realitio = realitio;
+        this.provider = provider;
+        this.rContract = new Contract(realitio, realitioIFace, provider);
+        this.cache = new LRU<string, ResultType>({max: 10000});
```
e1Ru1o, line 17: Note:
Usually when we have the blockchain interaction in a `handleBlock` we don't add a cache because there will not be repeated calls in the same block. Adding it is fine too thinking on cases where the scanner sends data twice or more times to the agent, but that is not the expected behavior.

SafeSnap-Agents/Cooldown-Alert/src/data.fetcher.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    let finalizeTS: number = await fetcher.getFinalizeTS(testBlockNumber, testQuestionId);
+    expect(finalizeTS).toStrictEqual(testFinalizeTS);
+
+    // clear mock to ensure cache usage
+    mockProvider.clear();
+
+    finalizeTS = await fetcher.getFinalizeTS(testBlockNumber, testQuestionId);
+    expect(finalizeTS).toStrictEqual(testFinalizeTS);
+  });
+});
```
e1Ru1o, line 45: Please add more cases in this tests in order to try more than one call

SafeSnap-Agents/Cooldown-Alert/src/agent.spec.ts
```diff
@@ -0,0 +1,290 @@
...
+
+    const blockEvent: BlockEvent = new TestBlockEvent().setNumber(testBlockNumber);
+
+    const findings: Finding[] = await handleBlock(blockEvent);
+
+    expect(findings).toStrictEqual([
+    ]);
+  });
+
+  it("should return multiple Findings from detecting different questions' cooldowns", async () => {
```
e1Ru1o, line 145:
```suggestion
  it("should return multiple Findings from detecting different question's cooldowns", async () => {
```

SafeSnap-Agents/Cooldown-Alert/src/agent.spec.ts
```diff
@@ -0,0 +1,290 @@
...
+  HandleBlock,
+  TransactionEvent,
+  BlockEvent,
+} from "forta-agent";
+import {
+  createAddress,
+  TestBlockEvent,
+  TestTransactionEvent,
+  encodeParameters,
+} from "forta-agent-tools";
```
e1Ru1o, line 15: check the `runBlock` tool from our module and consider using it in your tests

e1Ru1o (changes requested):

SafeSnap-Agents/Cooldown-Alert/src/agent.spec.ts
RCantu92: I have looked into it, and I don't think it'll work in my case. Saw the code is:
```
export async function runBlock(agent: Agent, block: BlockEvent, ...txns: TransactionEvent[]): Promise<Finding[]> {
  let findings: Finding[] = [];

  findings.push(...(await agent.handleBlock(block)));
  for (let tx of txns) findings.push(...(await agent.handleTransaction(tx)));

  return findings;
}
```
However, for my scenario, I need the handler to run `handleTransaction` before `handleBlock` because the txn handler is what populates the `questionIds` array, which the block handler then loops through.

SafeSnap-Agents/Cooldown-Alert/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "keyfile": "forta-agent keyfile",
+    "test": "jest"
+  },
+  "dependencies": {
+    "forta-agent": "^0.0.36"
+  },
+  "devDependencies": {
+    "@types/jest": "^27.0.1",
+    "@types/jest-when": "^3.5.0",
+    "@types/lru-cache": "^5.1.1",
```
e1Ru1o, line 27: remove this unused dependency

e1Ru1o (approved):

PR #151, opened 2022-02-17
answers submission agent

yasmine9:


SafeSnap-Agents/AnswersSubmission/README.md
```diff
@@ -0,0 +1,33 @@
...
+    - `questionId`: the question Id who's answer is submitted.
+    - `answer`: the submitted answer.
+    - `user`: address of the user who submitted the answer.
+    - `commitement`: set to `yes` if a commitment is submitted, `no` otherwise.
+
+- GNOSIS-3-2
+
+  - Fired when `LogAnswerReveal` event is emitted on the reality module oracle.
+
+- Severity is always set to "Info".
```
e1Ru1o, line 28: Fix this formatting here

SafeSnap-Agents/AnswersSubmission/src/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+export const ANSWERS_EVENTS_SIGNATURES = [
+  "event LogNewAnswer(bytes32 answer, bytes32 indexed question_id, bytes32 history_hash, address indexed user, uint256 bond, uint256 ts, bool is_commitment)",
+  "event LogAnswerReveal(bytes32 indexed question_id, address indexed user, bytes32 indexed answer_hash, bytes32 answer, uint256 nonce, uint256 bond)",
+];
+
+export const reality_abi = new Interface([
+  "function oracle() view returns (address)",
+  QUESTION_CREATED_SIGNATURE,
+]);
+export const oracle_abi = new Interface(ANSWERS_EVENTS_SIGNATURES);
```
e1Ru1o, line 17: I would like to suggest uppers for this vars too

SafeSnap-Agents/AnswersSubmission/src/oracle.fetcher.ts
```diff
@@ -0,0 +1,33 @@
...
+    const RealityContract = new Contract(
+      contract.toLowerCase(),
+      reality_abi,
+      this.provider
+    );
+    const totalPaymentReceived: Promise<string> = RealityContract.oracle({
+      blockTag: block,
+    });
+    this.cache.set(key, totalPaymentReceived);
+    return totalPaymentReceived;
```
e1Ru1o, line 31:
```suggestion
    const oracle: Promise<string> = RealityContract.oracle({
      blockTag: block,
    });
    this.cache.set(key, oracle);
    return oracle;
```

SafeSnap-Agents/AnswersSubmission/package.json
```diff
@@ -0,0 +1,35 @@
...
+  },
+  "devDependencies": {
+    "@types/jest": "^27.0.1",
+    "@types/nodemon": "^1.19.0",
+    "jest": "^27.0.6",
+    "nodemon": "^2.0.8",
+    "ts-jest": "^27.0.3",
+    "typescript": "^4.3.4",
+    "jest-when": "^3.5.1"
+  }
```
e1Ru1o, line 34: I guess you should add `@types/jest-when` also here

e1Ru1o (changes requested):

Kalzak (changes requested):

SafeSnap-Agents/AnswersSubmission/package.json
yasmine9: jest-when is no longer needed after using the `initialize` handler

SafeSnap-Agents/AnswersSubmission/README.md
```diff
@@ -0,0 +1,33 @@
...
+  - Type is always set to "Info".
+  - Metadata contains:
+    - `questionId`: the question Id who's answer is submitted.
+    - `answer`: the revealed answer.
+    - `user`: address of the user who revealed the answer.
```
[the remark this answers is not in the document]
yasmine9: fixed in the last commit

SafeSnap-Agents/AnswersSubmission/src/agent.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  beforeAll(async () => {
+    //init the mock fetcher
+    when(mockGetOracle)
+      .calledWith("latest", reality_module)
+      .mockReturnValue(oracle_address);
+    // init the agent ie. fetch the oracle.
+    handler = provideHandleTransaction("");
+    const initializer = initialize(reality_module, mockOracleFetcher as any);
+    await initializer();
```
e1Ru1o, line 92: You don't need to use the initialize logic here even more when the logic used in the initialize is tested in your fetcher tests. You can keep it that way if you want, but I prefer to see
```
handler = provideHandleTransaction(oracle_address);
```
and that will let you remove the `if(!_oracle) _oracle = oracle` line in your agent logic, that I think is there just for testing purposes.
If you want to keep using the initialize function here I think that you can return the oracle in the function, that is not the expected signature of the initialize function but it will not break anything. By doing that you will be able to get the oracle an create the handler after that.
My preference current preference is to not use the initialize logic in the tests.

e1Ru1o (approved):

SafeSnap-Agents/AnswersSubmission/src/agent.spec.ts
yasmine9: addressed in the last commit

Kalzak (approved):

PR #154, opened 2022-02-21
@impossible/staxmigrate

Kalzak:
# Staxmigrate Agent

## Types of changes

- [x] New Agent
- [ ] Fix an agent error

## New Agent PR readiness

- [x] Agent contains tests
- [x] Agent can recover from failed error call when they can fail (e.g. error are being catched on calls than can fail)


Impossible-Agent-Suite/staxmigrate/src/agent.ts
```diff
@@ -0,0 +1,63 @@
...
+  return async (tx: TransactionEvent): Promise<Finding[]> => {
+    // Setup the findings array
+    const findings: Finding[] = [];
+
+    // Get all times `Transfer` was emitted
+    const transfers = tx.filterLog(IF_ABI[0], address);
+
+    // Get all times `staxMigrate` was called
+    const migrations = tx.filterFunction(IF_ABI[1], address);
+
```
yasmine9, line 27: You can switch between the to declarations and add in between something like `if (!migrations) return findings;` to avoid filtering unneeded Transfers when there is no `staxMigrate` call in the transaction

Impossible-Agent-Suite/staxmigrate/src/agent.ts
```diff
@@ -0,0 +1,63 @@
...
+    // For each `staxMigrate` call
+    migrations.forEach((migration) => {
+      // For each `Transfer` event
+      transfers.forEach((transfer) => {
+        // Check if the `Transfer` event is a mint
+        if(transfer.args.from == ZERO_ADDR) {
+          // Get the amount of STAX tokens that are going 'in'
```
yasmine9, line 34: Do you think we need to check that the function caller `(tx.from)` ==  the address on `to` argument  of the Transfer event? I can't see a case where this won't be verified but I think it'd be better to precise it to avoid catching a wrong transfer event.

Impossible-Agent-Suite/staxmigrate/src/agent.ts
```diff
@@ -0,0 +1,63 @@
+import {
+  Finding,
+  FindingType,
+  FindingSeverity,
+  TransactionEvent,
+  HandleTransaction
+} from 'forta-agent';
+
+let IF_ADDR = '0xb0e1fc65c1a741b4662b813eb787d369b8614af1';
```
yasmine9, line 9: variable not used in the agent

yasmine9: should be used in the export default

Impossible-Agent-Suite/staxmigrate/src/agent.ts
```diff
@@ -0,0 +1,63 @@
...
+let ZERO_ADDR = '0x0000000000000000000000000000000000000000';
```
yasmine9, line 10:
```suggestion
const IF_ADDR = '0xb0e1fc65c1a741b4662b813eb787d369b8614af1';
const ZERO_ADDR = '0x0000000000000000000000000000000000000000';
```

yasmine9 (changes requested):

Impossible-Agent-Suite/staxmigrate/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Description
+
+This agent detects when Impossible Finance tokens are minted through the function `staxMigrate` where the amount of `STAX` tokens input are not the same as the amount of `IF` tokens output
+
+## Supported Chains
+
+- Ethereum
+- BSC
+
```
yasmine9, line 11: based on the documentation, Ploygon is a supported chain as well

Impossible-Agent-Suite/staxmigrate/src/agent.ts
```diff
@@ -0,0 +1,63 @@
...
+            );
+          }
+        }
+      });
+    });
+
+    // Return findings
+    return findings;
+  }
+}
```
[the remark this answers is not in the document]
Kalzak: Oops! I didn't have a live TX to test with so I never ended up having the agent run through that way. Fixed now!

Impossible-Agent-Suite/staxmigrate/src/agent.ts
Kalzak: Good idea, that has been implemented.

Impossible-Agent-Suite/staxmigrate/src/agent.ts
Kalzak: I gave it some thought during development and it's very likely that every single call to this agent will be from an EOA, but there is a _tiny_ chance that a smart contract may be the caller, in which case this agent wouldn't detect anything.
Given that we're looking for unusual minting behaviour, if an attacker were to exploit the minting process a smart contract as the caller would be more likely than an EOA.

It's very unlikely that the agent would catch a wrong transfer event because we are only looking for a `Transfer` from the zero address which means it's a mint. Catching a wrong mint is very unlikely since calling `staxMigrate` calls `_mint` exactly once.

Kalzak: I may be missing something though, I'd like to hear your thoughts on this

Impossible-Agent-Suite/staxmigrate/src/agent.ts
Kalzak: `IF_ADDR` is now used in the `export` in `agent.ts`.

Impossible-Agent-Suite/staxmigrate/src/agent.ts
Kalzak: This change has been implemented.

Impossible-Agent-Suite/staxmigrate/README.md
Kalzak: I'm unsure if I should event state that we support the Ethereum chain now. After reviewing Lazaro's PRs he states that BSC is the only supported chain in his READMEs. Thoughts @e1Ru1o ?

Impossible-Agent-Suite/staxmigrate/src/agent.ts
yasmine9: Very good analysis, that makes sense 
I agree with the above

Impossible-Agent-Suite/staxmigrate/README.md
e1Ru1o: Yes use only BSC

Kalzak: Addressed in latest commit.

yasmine9 (approved):

Impossible-Agent-Suite/staxmigrate/src/agent.ts
```diff
@@ -0,0 +1,70 @@
...
+      migrations.forEach((migration) => {
+        // For each `Transfer` event
+        transfers.forEach((transfer) => {
+          // Check if the `Transfer` event is a mint
+          if(transfer.args.from == ZERO_ADDR) {
+            // Get the amount of STAX tokens that are going 'in'
+            const staxAmountIn = migration.args.amount.toString();
+            const ifAmountOut = transfer.args.value.toString();
+            // If `staxAmountIn` and `ifAmountOut` are different 
+            if(staxAmountIn != ifAmountOut) {
```
e1Ru1o, line 38: This will not work for txns with multiple migrations. Lets say 2 valid migrations occur with different amounts, this logic will generate two findings.
You should do something like a two pointers algorithm to find a match for each `staxMigrate` call

e1Ru1o (changes requested):

e1Ru1o (changes requested):
Please add a test with multiple findings to ensure that the problem is fixed. Agent logic LGTM.

yasmine9 (commented):
I agree with Lazaro's suggestion to enrich the tests. Other than that, the update looks good to me

e1Ru1o (approved):

PR #168, opened 2022-03-11
@paraswap/delegate calls

Kalzak:
# Delegated Function Call Agent

## Types of changes

- [x] New Agent
- [ ] Fix an agent error

## New Agent PR readiness

- [x] Agent contains tests
- [x] Agent can recover from failed error call when they can fail (e.g. error are being catched on calls than can fail)


Paraswap-Agents/delegate-calls/src/agent.ts
```diff
@@ -0,0 +1,96 @@
...
+    });
+
+    // Check if any contracts have had roles revoked and update `cache` if needed
+    const roleRevokes = tx.filterLog(PARA_ABI[2], address);
+    roleRevokes.forEach((roleRevoke) => {
+      // If ROUTER_ROLE was removed from an account that is inside the cache then update the cache
+      if (roleRevoke.args.role == ROUTER_ROLE && cache.has(roleRevoke.args.account)) {
+        cache.set(roleRevoke.args.account, false);
+      }
+    });
```
yasmine9, line 74: Can't we avoid these updates with every tx by specifying the block number in `hasRole` call? 
it'll cause updates to be done with each blockEvent

Paraswap-Agents/delegate-calls/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "delegated-function-call-agent",
+  "version": "0.0.1",
+  "description": "Detects delecated calls done on the Paraswap protocol",
```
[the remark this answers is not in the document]
Kalzak: Fixed in following commit.

Paraswap-Agents/delegate-calls/README.md
```diff
@@ -0,0 +1,42 @@
...
+    - This means that the contract is foreign to the protocol and unknown code is being executed
+  - Severity is set to "High"
+  - Type is set to "Exploit"
+  - Metadata:
+    - `logicContract`: The contract called by the `AugustusSwapper` contract through a `delegatecall`
+
+- PARASWAP-3-3
+  - Fired when `AugustusSwapper` makes a delegated call to a contract and it cannot be verified if the contract has the role `ROUTER_ROLE`
+    - This will only occur if the `ethers` call to `AugustusSwapper.hasRole()` fails. This is very unlikely to ever fire however in the case that a delegated call is made while the agent's endpoint is having issues, the agent does not know if the contract is trusted (has role `ROUTER_ROLE`) or is not trusted (doesn't have role `ROUTER_ROLE`). It would then be recommended for the Paraswap team to manually verify if the `logicContract` has the role `ROUTER_ROLE` or not.
+  - Severity is set to "Info"
```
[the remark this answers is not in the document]
Kalzak: Fixed in following commit.

Paraswap-Agents/delegate-calls/src/agent.spec.ts
```diff
@@ -0,0 +1,223 @@
...
+  let contract: ethers.utils.Interface;
+  let mockCheckRouterRole: any;
+
+  beforeEach(() => {
+    // Setup the mock `checkRouterRole` function
+    mockCheckRouterRole = jest.fn();
+    // Setup the handler to use the mocked `checkRouterRole` function
+    handler = provideHandleTransaction(AUGUSTUS_ADDR, mockCheckRouterRole);
+    // Setup an ethers interface to be used when creating function calls in the test transactions
+    contract = new ethers.utils.Interface(["function exampleFunctionCall(bool) returns (bool)"]);
```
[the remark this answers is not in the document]
Kalzak: Fixed in following commit.

Paraswap-Agents/delegate-calls/src/agent.spec.ts
```diff
@@ -0,0 +1,223 @@
...
+    });
+
+    // Run the handler on the test transaction
+    const findings: Finding[] = await handler(tx);
+    // Check if findings contain expected results
+    expect(findings).toStrictEqual([]);
+  });
+
+  it("ignores delegated calls from contracts not in the `addresses` array", async () => {
+    // Create a transaction with a delegated call from a external address to a external address
```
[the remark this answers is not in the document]
Kalzak: I forgot to update a test name after changing the logic in my `agent.ts`.
`addresses` doesn't exist in my current implementation. 
Fixed in following commit.

Paraswap-Agents/delegate-calls/src/agent.ts
Kalzak: I've added specifying the block number to the call to `hasRole` in the function `checkRouterRole` to prevent web3 calls to a future block while the agent is still working on the previous block.

As for you suggestion to change the way we keep up to date with role changes, the current implementation and your suggestion both have positives and negatives.

New suggestion:
- Removes need to run `tx.filterLog` for on role grant and revoke events
- Adds need to make a web3 call to `AugustusSwapper.hasRole` on every delegated call

Current implementation:
- Needs to run `tx.filterLog` twice
- Number of web3 calls are reduced since the roles and their associated addresses are stored in the LRU cache.

@e1Ru1o 
Yasmine and I discussed this in a call and aren't sure which approach is better. Thoughts?

Paraswap-Agents/delegate-calls/src/agent.ts
```diff
@@ -0,0 +1,100 @@
...
+      tx.traces.map(async (trace) => {
+        // If the calltype is "delegatecall" and the from address is `address`
+        if (trace.action.callType == "delegatecall" && trace.action.from == address) {
+          // Check if the logic contract for the delegatecall has the role ` ROUTER_ROLE`
+          const hasRouterRole = await checkRouterRole(address, trace.action.to, tx.blockNumber);
+          // Create a finding
+          findings.push(createFinding(hasRouterRole, trace.action.to));
+        }
+      })
+    );
```
RCantu92, line 91: Since the `AugustusSwapper` contract is a proxy that `delegatecall`s to other contracts for most of its functionality, I think it would be more helpful to be selective under what conditions Findings are created. Specifically when the logic contract **has** the `ROUTER_ROLE`.

Perhaps filtering for specific function calls in the logic contract (e.g. `multiswap` and `megaswap` per the Paraswap Notion page)? To be even more selective, perhaps, using `multiswap` and `megaswap` as examples, only when `toAmount` passes a certain threshold.

RCantu92 (changes requested):
I think this is a fantastic job in approaching `delegatecall`s with an agent. I also think it is helpful to detect `delegatecall`s to contracts that don't have `ROUTER_ROLE`.

However, I think we need to adjust our approach with detecting `delegatecall`s when the logic contract **does** have `ROUTER_ROLE`.

Currently, the agent detects all `delegatecall`s, and with `AgustusSwapper` being a proxy contract, that would be almost too many. I think it would be a good idea  to have a criteria to filter for more specific function calls to logic contracts, in cases when the logic contract **does** have the `ROUTER_ROLE`.

Paraswap-Agents/delegate-calls/src/agent.ts
```diff
@@ -0,0 +1,100 @@
+import {
+  Finding,
+  FindingType,
+  FindingSeverity,
```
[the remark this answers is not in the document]
Kalzak: Addressed in following commit.

Paraswap-Agents/delegate-calls/src/agent.ts
```diff
@@ -0,0 +1,100 @@
...
+      // If there is as error set `isRouterRole` to undefined
+      // A special finding to indicate that the on-chain call failed is generated when `isRouterRole` is undefined
+      // This is not saved to cache in case it was a short term issue
+      isRouterRole = undefined;
+    }
+  }
+  return isRouterRole;
+};
+
+export const provideHandleTransaction = (address: string, checkRouterRole: any): HandleTransaction => {
```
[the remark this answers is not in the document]
Kalzak: Addressed in following commit.

Paraswap-Agents/delegate-calls/src/agent.ts
```diff
@@ -0,0 +1,100 @@
...
+): Promise<boolean | undefined> => {
+  let isRouterRole;
+  // If the contract exists in the cache then return data
+  if (cache.has(logicAddr)) {
+    isRouterRole = cache.get(logicAddr);
+  // Otherwise the cache does not contain data for `logicAddr`
+  } else {
+    try {
+      // Check on-chain data to see if `logicAddr` has the correct role
+      const contractInterface = new ethers.Contract(paraAddr, PARA_ABI, getEthersProvider());
```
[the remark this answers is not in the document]
Kalzak: Addressed in following commit.

Paraswap-Agents/delegate-calls/src/finding.ts
```diff
@@ -0,0 +1,41 @@
+import { Finding, FindingType, FindingSeverity, TransactionEvent, HandleTransaction } from "forta-agent";
```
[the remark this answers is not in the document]
Kalzak: Addressed in following commit.

Paraswap-Agents/delegate-calls/README.md
```diff
@@ -0,0 +1,42 @@
+# Delegated Function Call Agent
+
+## Description
+
+This agent detects delagated calls by the contract `AugustusSwapper` (see [here](https://developers.paraswap.network/smart-contracts#augustusswapper) for more details)
+
+## Supported Chains
+
+- Ethereum
```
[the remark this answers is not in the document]
Kalzak: Addressed in following commit.

Paraswap-Agents/delegate-calls/src/agent.spec.ts
```diff
@@ -0,0 +1,226 @@
+import { Finding, FindingType, FindingSeverity, TransactionEvent, HandleTransaction } from "forta-agent";
```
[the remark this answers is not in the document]
Kalzak: Addressed in following commit.

Paraswap-Agents/delegate-calls/src/agent.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+
+import { TestTransactionEvent, createAddress } from "forta-agent-tools/lib/tests";
+
+import { ethers } from "ethers";
+
+import { provideHandleTransaction, PARA_ABI } from "./agent";
```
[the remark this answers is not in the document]
Kalzak: Addressed in following commit.

Paraswap-Agents/delegate-calls/src/agent.ts
e1Ru1o: I understand what you are doing and it seems like a great effort to improve performance. I guess that you are right and your implementation has a better performance because the network calls are more expensive than the filters.

So my only concern here is that if for some reason the agent miss a txn or analyze some txns out of order I guess false positive/negative can happen with the data you have in the cache since it will not be updated correctly. In other words your approach is dependent of the order in where the txns are processed. Forta team tell us before that this is not a problem and we can assume the txns are going to be processed in order, but we always try to be careful with that due to the asynchronous behavior of the agents. 

I just checked the sdk [here](https://github.com/forta-protocol/forta-agent-sdk/blob/master/cli/commands/run/run.live.ts#L8) and as Forta said we don't need to worry about that, blocks are being processed one by one as well as txns. 

Said the above I think you can keep your approach. You are setting the address to `true` or `false` accordingly to the events and in the cases you don't have info in the cache you are fetching the data from the blockchain so everything looks good to me right now.

Paraswap-Agents/delegate-calls/src/agent.ts
e1Ru1o: That is right, we are going to deliver detection for only `multiswap` and `megaswap` calls + the `ROUTER_ROLE` monitoring. So please add the findings for each time that functions are called.

Paraswap-Agents/delegate-calls/README.md
```diff
@@ -0,0 +1,32 @@
...
+    - `logicContract`: The contract called by the `AugustusSwapper` contract through a `delegatecall`
+
+## Test Data
+
+The agent behaviour can be verified with the following transactions:
+- On Paraswap protocol (Ethereum):
+  - `0x161929d7651f1e231ed93d42a620a85aefba7ff154a34eb388bd1179d6192ce2` (delegatecall to `multiSwap`/`megaSwap` function, finding generated)
+  - `0x94c0a7cd60587e03dfd92b5ee200d16cef7834c10267fe12dd178f6659f1c702` (delegatecall to a non `multiSwap`/`megaSwap` function, no finding generated)
+- On testing protocol (Ropsten):
+  - Not yet implemented
```
yasmine9, line 32: Since we're delivering the agents before developing the PoC, I think this should be deleted for now

Paraswap-Agents/delegate-calls/src/agent.ts
```diff
@@ -0,0 +1,51 @@
...
+import { ethers } from "ethers";
+
+import { createFinding } from "./finding";
+
+// The address of the Paraswap contract (the contract's name is "AugustusSwapper")
+const AUGUSTUS_ADDR = "0xdef171fe48cf0115b1d80b88dc8eab59176fee57";
+
+// Function signatures for `simpleSwap` and `simpleBuy`
+const simpleSwapFuncSigs = ["0x54e3f31b", "0x2298207a"];
+
```
yasmine9, line 19: remove unused variable

Paraswap-Agents/delegate-calls/README.md
```diff
@@ -0,0 +1,32 @@
...
+- Ropsten
+
+## Alerts
+
+- PARASWAP-3
+  - Fired when `AugustusSwapper` makes a delegated call to `MultiPath.multiSwap` or `MultiPath.megaSwap`
+  - Severity is set to "Info"
+  - Type is set to "Info"
+  - Metadata:
+    - `logicContract`: The contract called by the `AugustusSwapper` contract through a `delegatecall`
```
0xtaf, line 23: metadata should include `multiPathAddress` instead of `logicContract`

Paraswap-Agents/delegate-calls/src/agent.ts
```diff
@@ -0,0 +1,51 @@
...
+  Finding,
+  FindingSeverity,
+  FindingType,
+  TransactionEvent,
+  HandleTransaction,
+  getEthersProvider,
+} from "forta-agent";
+
+import { ethers } from "ethers";
+
```
yasmine9, line 11: Please removed unused imports: `getEthersProvider ` and `ethers`

Paraswap-Agents/delegate-calls/src/agent.ts
```diff
@@ -0,0 +1,51 @@
...
+    await Promise.all(
+      tx.traces.map(async (trace) => {
+        // If the calltype is "delegatecall" and the from address is `address`
+        if (trace.action.callType == "delegatecall" && trace.action.from == paraAddr) {
+          // Get the function signature used in the call
+          const funcSig = trace.action.input.substring(0,10);
+          // If the function signature matches `multiSwap` or `megaSwap`
+          if(multiPathFuncSigs.includes(funcSig)) {
+            // Create a finding
+            findings.push(createFinding(trace.action.to));
```
yasmine9, line 38: Since we got the info in `funcSig` , I would suggest including the swap type ( `multiSwap` or `megaSwap`) in the finding. For example,  you can pass `funcSig` to `createFinding` and do the check inside that one.

Paraswap-Agents/delegate-calls/src/agent.ts
```diff
@@ -0,0 +1,51 @@
+import {
+  Finding,
+  FindingSeverity,
+  FindingType,
+  TransactionEvent,
+  HandleTransaction,
+  getEthersProvider,
+} from "forta-agent";
+
+import { ethers } from "ethers";
```
0xtaf, line 10: `FindingSeverity`, `FindingType`, `getEthersProvider` and `ethers` are not used in this file.

Paraswap-Agents/delegate-calls/src/agent.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { createFinding } from "./finding";
+
+// The address of the Paraswap contract (the contract's name is "AugustusSwapper")
+const AUGUSTUS_ADDR = "0xdef171fe48cf0115b1d80b88dc8eab59176fee57";
+
+// Function signatures for `simpleSwap` and `simpleBuy`
+const simpleSwapFuncSigs = ["0x54e3f31b", "0x2298207a"];
```
0xtaf, line 18: `simpleSwapFuncSigs` is declared but not used.

yasmine9 (changes requested):
Some unused imports and variables should be deleted. I suggest to include the swap Type in the finding description or as a metadata. Other than that, LGTM 👍

Paraswap-Agents/delegate-calls/src/finding.ts
```diff
@@ -0,0 +1,15 @@
...
+export const createFinding = (multiPathAddress: string) => {
+  return Finding.fromObject({
+    name: "MultiPath DelegateCall",
+    description: "A delegated call has been made to the MultiPath contract",
+    alertId: "PARASWAP-3",
+    severity: FindingSeverity.Info,
+    type: FindingType.Info,
+    protocol: "Paraswap",
+    metadata: {
+      multiPathAddress: multiPathAddress
```
0xtaf, line 12: this is not a deal but just for convenience, since the and the value have the same name, it could also be written like this:
```
metadata: {
  multiPathAddress
}
```

0xtaf (approved):

Paraswap-Agents/delegate-calls/src/agent.ts
Kalzak: Great suggestion, this has been implemented in the latest commit.

yasmine9 (approved):

Paraswap-Agents/delegate-calls/src/agent.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+        input: "0xa94e78ef",
+        output: "0x0",
+      })
+      .addTraces({
+        to: EXTERNAL_ADDR_2,
+        from: AUGUSTUS_ADDR,
+        // Manually place the function signature due to function argument complexity
+        input: "0x46c67b6d",
+        output: "0x0",
+      })
```
RCantu92, line 137:
```suggestion
      .addTraces({
        to: EXTERNAL_ADDR_1,
        from: AUGUSTUS_ADDR,
        // Manually place the function signature due to function argument complexity
        input: "0xa94e78ef",
        output: "0x0",
      })
      .addTraces({
        to: EXTERNAL_ADDR_1,
        from: AUGUSTUS_ADDR,
        // Manually place the function signature due to function argument complexity
        input: "0x46c67b6d",
        output: "0x0",
      })
      .addTraces({
        to: EXTERNAL_ADDR_2,
        from: AUGUSTUS_ADDR,
        // Manually place the function signature due to function argument complexity
        input: "0xa94e78ef",
        output: "0x0",
      })
      .addTraces({
        to: EXTERNAL_ADDR_2,
        from: AUGUSTUS_ADDR,
        // Manually place the function signature due to function argument complexity
        input: "0x46c67b6d",
        output: "0x0",
      })
```
Small suggestion to have tried different `to` contracts per function signature, so as to only keep the function signatures the same per set of `Traces`.

Paraswap-Agents/delegate-calls/src/agent.ts
```diff
@@ -0,0 +1,43 @@
...
+  HandleTransaction,
+} from "forta-agent";
+
+import { createFinding } from "./finding";
+
+// The address of the Paraswap contract (the contract's name is "AugustusSwapper")
+const AUGUSTUS_ADDR = "0xdef171fe48cf0115b1d80b88dc8eab59176fee57";
+
+// Function signatures for `multiSwap` and `megaSwap`
+const multiPathFuncSigs = ["0xa94e78ef", "0x46c67b6d"];
```
RCantu92, line 13:
```suggestion
const MULTIPATH_FUNC_SIGS = ["0xa94e78ef", "0x46c67b6d"];
```
Small suggestion to make variable all caps since it is a similar variable to `AUGUSTUS_ADDR`.

Paraswap-Agents/delegate-calls/README.md
```diff
@@ -0,0 +1,35 @@
...
+    - `multiPathAddress`: The contract called by the `AugustusSwapper` contract through a `delegatecall`
+
+## Test Data
+
+The agent behaviour can be verified with the following transactions:
+- On Paraswap protocol (Ethereum):
+  - `0x161929d7651f1e231ed93d42a620a85aefba7ff154a34eb388bd1179d6192ce2`
+    - (delegatecall to `multiSwap`, finding generated)
+  - `0x9d5ed5e3146461f0fc99b47ccd7f5694fae1080c255de9ce8ae3bc4464957632`
+    - (delegatecall to `megaSwap`, finding generated)
```
RCantu92, line 33: I ran the txns and couldn't get a Finding generated for either. Even confirmed my `forta.config.json` was properly set up and still nothing. (I checked Infura, and the requests did go through.)

RCantu92 (changes requested):
I couldn't generate Findings from the mainnet transactions provided. I checked my `forta.config` to make sure it was set up for Ethereum mainnet, and it looks correct to me. Also checked Infura, and the request went through.

However, if some else can confirm these transactions generate Findings for them and it's an issue only with me, I will approve.

RCantu92 (approved):
Correctly updated my `traceRpcUrl` in the `forta.config` and now have the Findings. 👍

PR #169, opened 2022-03-11
@paraswap/admin operations

yasmine9:
# Agent name here

## Types of changes

- [x] New Agent
- [ ] Fix an agent error

## New Agent PR readiness

- [ ] The `name` and `description` fields of `package.json` describe the agent properly
- [ ] The agent README is updated
- [ ] The agent README sufficiently describes what the agent does
- [ ] All the alerts that the agent emits are listed in the README
- [ ] Code is easy to follow and/or properly comment
- [ ] Agent contains tests
- [ ] All tests pass when running `npm tests`
- [ ] The code do what the README describes
- [ ] Findings `protocol` is specified (if the agent is for an specific one)
- [ ] the agent follows Forta recommended [best practices](https://docs.forta.network/en/latest/best-practices/)

## Comments (optional)

### DIFF
--- a/Paraswap-Agents/AdminOperations/README.md
+++ b/Paraswap-Agents/AdminOperations/README.md
@@ -7,6 +7,11 @@ This agent detects admin operations on AugustusSwapper Contract.
 ## Supported Chains
 
 - Ethereum
+- Polygon
+- BSC
+- Avalanche
+- Fantom
+- Ropsten
 
 ## Alerts
 
--- a/Paraswap-Agents/AdminOperations/README.md
+++ b/Paraswap-Agents/AdminOperations/README.md
@@ -1,0 +1,58 @@
+# Admin Operations Agent
+
+## Description
+
+This agent detects admin operations on AugustusSwapper Contract.
+
+## Supported Chains
+
+- Ethereum
+
+## Alerts
+
+- PARASWAP-1-1
+
+  - Fired when the Router or Adapter has been initialized.
+  - Severity is always set to "Info".
+  - Type is always set to "Info".
+  - Metadata fields:
+    - `address`: Address of the initialized router/adapter.
+
+- PARASWAP-1-2
+
+  - Fired when the tokens were transfers through calls to `transferTokens`.
+  - Severity is always set to "Info".
+  - Type is always set to "Info".
+  - Metadata fields:
+    - `token`: Address of the transfered token.
+    - `destination`: Address to which the tokens were transfered.
+    - `amount`: Amount of transfered tokens.
+
+- PARASWAP-1-3
+
+  - Fired when the router implementation was upgraded through calls to `setImplementation`.
+  - Severity is always set to "Info".
+  - Type is always set to "Info".
+  - Metadata fields:
+    - `selector`: Selector of the new implementation.
+    - `implementation`: Address of the new router implementation.
+
+- PARASWAP-1-4
+
+  - Fired when a new partner was registred through calls to `registerPartner`.
+  - Severity is always set to "Info".
+  - Type is always set to "Info".
+  - Metadata fields:
+    - `partner`: Address of the new partner.
+    - `partnerShare`: The partner Share.
+    - `noPositiveSlippage`: Set to `true` if positive slippage should not be taken into account. `false` otherwise.
+    - `positiveSlippageToUser`: Set to `true` if positive slippage should go to the user. `false` otherwise.
+    - `feePercent`: Partner fee Percent.
+
+- PARASWAP-1-5
+
+  - Fired when the Fee Wallet address was changed through calls to `setFeeWallet`.
+  - Severity is always set to "Info".
+  - Type is always set to "Info".
+  - Metadata fields:
+    - `new_address`: Address of the new Fee Wallet.
--- a/Paraswap-Agents/AdminOperations/src/agent.spec.ts
+++ b/Paraswap-Agents/AdminOperations/src/agent.spec.ts
@@ -1,0 +1,240 @@
+import { BigNumber } from "ethers";
+import { Interface } from "ethers/lib/utils";
+import { Finding, FindingType, FindingSeverity, TransactionEvent, HandleTransaction } from "forta-agent";
+import { createAddress, TestTransactionEvent } from "forta-agent-tools/lib/tests";
+import { provideHandleTransaction } from "./agent";
+import { SWAPPER_IFACE } from "./utils";
+
+// AugustusSwapper address
+const SWAPPER_ADDR = createAddress("0xa1");
+// Address used to initiate transactions
+const USER_ADDR = createAddress("0xaa");
+// Irrelevent events/ functions used for tests
+const IRRELEVENT_IFACE = new Interface(["event wrongsig()", "function wrongFunction()"]);
+
+const TEST_DATA = [
+  [createAddress("0xb1"), createAddress("0xc1"), BigNumber.from(10)], // transferTokens args
+  [createAddress("0xb2"), createAddress("0xc2"), BigNumber.from(80)], // transferTokens args
+  [createAddress("0xb3"), createAddress("0xc3"), BigNumber.from(90)], // transferTokens args
+  [USER_ADDR, createAddress("0xc1"), BigNumber.from(10)], // Transfer event args
+  [USER_ADDR, createAddress("0xc2"), BigNumber.from(80)], // Transfer event args
+  ["0xafde1234", createAddress("0xd1")], // setImplementation args
+  [createAddress("0xd3")], // setFeeWallet args
+  [createAddress("0xd4")], // RouterInitialized args
+  [createAddress("0xd5")], // AdapterInitialized args
+];
+
+// function to generate test findings
+const createFinding = (operationName: string, args: any[]) => {
+  switch (operationName) {
+    case "transferTokens":
+      return Finding.fromObject({
+        name: `Admin operation detected: tokens were transfered`,
+        description: `${operationName} function was called and a Transfer event was emitted in AugustusSwapper contract`,
+        alertId: "PARASWAP-1-2",
+        severity: FindingSeverity.Info,
+        type: FindingType.Info,
+        protocol: "Paraswap",
+        metadata: {
+          token: args[0].toLowerCase(),
+          destination: args[1].toLowerCase(),
+          amount: args[2].toString(),
+        },
+      });
+
+    case "setImplementation":
+      return Finding.fromObject({
+        name: `Admin operation detected: Implementation upgraded`,
+        description: `${operationName} function was called in AugustusSwapper contract`,
+        alertId: "PARASWAP-1-3",
+        severity: FindingSeverity.Info,
+        type: FindingType.Info,
+        protocol: "Paraswap",
+        metadata: {
+          selector: args[0].toLowerCase(),
+          implementation: args[1].toLowerCase(),
+        },
+      });
+
+    case "setFeeWallet":
+      return Finding.fromObject({
+        name: `Admin operation detected: Fee Wallet address changed`,
+        description: `${operationName} function was called in AugustusSwapper contract`,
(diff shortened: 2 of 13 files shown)


Paraswap-Agents/AdminOperations/README.md
```diff
@@ -0,0 +1,63 @@
...
+- BSC
+- Avalanche
+- Fantom
+- Ropsten
+
+## Alerts
+
+- PARASWAP-1-1
+
+  - Fired when the Router or Adapter has been initialized.
```
RCantu92, line 20:
```suggestion
  - Fired when the `Router` or `Adapter` has been initialized.
```

Paraswap-Agents/AdminOperations/README.md
```diff
@@ -0,0 +1,63 @@
...
+    - `implementation`: Address of the new router implementation.
+
+- PARASWAP-1-4
+
+  - Fired when a new partner was registred through calls to `registerPartner`.
+  - Severity is always set to "Info".
+  - Type is always set to "Info".
+  - Metadata fields:
+    - `partner`: Address of the new partner.
+    - `partnerShare`: The partner Share.
```
RCantu92, line 52:
```suggestion
    - `partnerShare`: The partner share.
```

Paraswap-Agents/AdminOperations/src/agent.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+  [USER_ADDR, createAddress("0xc1"), BigNumber.from(10)], // Transfer event args
+  [USER_ADDR, createAddress("0xc2"), BigNumber.from(80)], // Transfer event args
+  ["0xafde1234", createAddress("0xd1")], // setImplementation args
+  [createAddress("0xd3")], // setFeeWallet args
+  [createAddress("0xd4")], // RouterInitialized args
+  [createAddress("0xd5")], // AdapterInitialized args
+];
+
+// function to generate test findings
+const createFinding = (operationName: string, args: any[]) => {
```
RCantu92, line 28: Missing option to create a Finding with `alertId` of `PARASWAP-1-4` (`registerPartner`).

Paraswap-Agents/AdminOperations/src/agent.ts
```diff
@@ -0,0 +1,62 @@
...
+            transfer.args.to.toLowerCase() == call.args.destination.toLowerCase() &&
+            BigNumber.from(transfer.args.value).eq(BigNumber.from(call.args.amount))
+          ) {
+            findings.push(createFunctionFinding(call));
+            transferLogs.splice(transferLogs.indexOf(transfer), 1);
+            break;
+          }
+        }
+        // generate findings for the other function calls
+      } else findings.push(createFunctionFinding(call));
```
RCantu92, line 53: As it currently is, we are only creating alerts when the `transferTokens` function is called **AND** the `Transfer` event is emitted.

However, in the `transferTokens` function, there is an [`if` statement](https://etherscan.io/address/0xDEF171Fe48CF0115B1d80b88dc8eAB59176FEe57#code#F1#L1610) that checks whether it is an ETH transfer or an ERC-20 transfer.

Because of this `if` statement, we need to also create alerts when there is an ETH transfer, which is when the `transferTokens` function is called **but does not** emit a `Transfer` event.

RCantu92 (changes requested):
The agent logic needs to account for instances where the `transferTokens` function is called but does not emit a `Transfer` event. The times it does this is when it is a ETH transfer.

Also add the necessary test(s).

Paraswap-Agents/AdminOperations/src/findings.ts
```diff
@@ -0,0 +1,81 @@
...
+        alertId: "PARASWAP-1-4",
+        severity: FindingSeverity.Info,
+        type: FindingType.Info,
+        protocol: "Paraswap",
+        metadata: {
+          partner: call.args.partner.toLowerCase(),
+          partnerShare: call.args._partnerShare.toLowerCase(),
+          noPositiveSlippage: call.args._noPositiveSlippage,
+          positiveSlippageToUser: call.args._positiveSlippageToUser,
+          feePercent: call.args._feePercent,
```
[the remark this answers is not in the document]
yasmine9: Yes it's better for readability, agent updated.

Paraswap-Agents/AdminOperations/src/agent.ts
yasmine9: Oh I have missed that ETH transfer part using the low level `call` method. Nice catch 👍 
Agent updated so that all `transferTokens` calls generate findings. The `Transfer` event is no longer needed.

Paraswap-Agents/AdminOperations/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+  - Fired when the `Router` or `Adapter` has been initialized.
+  - Severity is always set to "Info".
+  - Type is always set to "Info".
+  - Metadata fields:
+    - `address`: Address of the initialized router/adapter.
+
+- PARASWAP-1-2
+
+  - Fired when the tokens were transfers through calls to `transferTokens`.
```
Kalzak, line 28: Fired when tokens are transferred through a call to `transferTokens`.

Kalzak (commented):
formatting/readability recommendations

Paraswap-Agents/AdminOperations/README.md
```diff
@@ -0,0 +1,63 @@
...
+  - Severity is always set to "Info".
+  - Type is always set to "Info".
+  - Metadata fields:
+    - `token`: Address of the transfered token.
+    - `destination`: Address to which the tokens were transfered.
+    - `amount`: Amount of transfered tokens.
+
+- PARASWAP-1-3
+
+  - Fired when the router implementation was upgraded through calls to `setImplementation`.
```
Kalzak, line 38: The `AugustusSwapper` smart contract supports multiple router implementations. 
It is possible to either
- Upgrade a router implementation by overwriting it's existing selector
- Add a new router implementation by using a unique selector that doesn't currently exist in the array `selectorVsRouter[]`

Adding checks to differentiate between the two would make the agent need to call to on-chain data which would slow the agent down. May be worth asking the client what they would prefer for this finding.

Paraswap-Agents/AdminOperations/src/findings.ts
```diff
@@ -0,0 +1,81 @@
...
+        metadata: {
+          partner: call.args.partner.toLowerCase(),
+          partnerShare: call.args._partnerShare.toString(),
+          noPositiveSlippage: call.args._noPositiveSlippage,
+          positiveSlippageToUser: call.args._positiveSlippageToUser,
+          feePercent: call.args._feePercent.toString(),
+        },
+      });
+
+    default:
```
Kalzak, line 67: I know that we're expecting `default` to always be `"setFeeWallet"` so using default is fine, but it's better to be explicit in this case rather than using `default` and then adding a comment.

Kalzak (changes requested):

Paraswap-Agents/AdminOperations/src/agent.spec.ts
```diff
@@ -0,0 +1,235 @@
...
+      // Add log of `AdapterInitialized` event
+      .addAnonymousEventLog(SWAPPER_ADDR, log1.data, ...log1.topics);
+
+    const findings = await handler(tx);
+    expect(findings).toStrictEqual([
+      createFinding("AdapterInitialized", TEST_DATA[6]),
+      createFinding("setImplementation", TEST_DATA[3]),
+      createFinding("setFeeWallet", TEST_DATA[4]),
+      createFinding("registerPartner", TEST_DATA[7]),
+    ]);
```
RCantu92, line 205: Would be good to include the `RouterInitialized` event emission as well. All the others seem to be covered by the tests.

Paraswap-Agents/AdminOperations/src/findings.ts
```diff
@@ -0,0 +1,81 @@
...
+    severity: FindingSeverity.Info,
+    type: FindingType.Info,
+    protocol: "Paraswap",
+    metadata: {
+      address: log.args.router ? log.args.router.toLowerCase() : log.args.adapter.toLowerCase(),
+    },
+  });
+};
+
+export const createFunctionFinding = (call: TransactionDescription) => {
```
RCantu92, line 19: I think it would be good to have an exclusive Finding for when it is an ETH transfer. With this current `transferTokens` Finding, we wouldn't know it is an ETH transfer except for the `token` in the `metadata` would be `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE`.

Perhaps an ETH transfer Finding could be:
```
return Finding.fromObject({
        name: `Admin operation detected: ETH was transferred`,
        description: `${call.name} function was called in AugustusSwapper contract`,
        alertId: "PARASWAP-1-3", // adjust the rest
        severity: FindingSeverity.Info,
        type: FindingType.Info,
        protocol: "Paraswap",
        metadata: {
          token: "ETH",
          destination: call.args.destination.toLowerCase(),
          amount: call.args.amount.toString(),
        },
      });
```
Then adding the appropriate unit test.

RCantu92 (changes requested):
Request to add a separate Finding for when the `transferTokens` function is called for an ETH transfer.

Paraswap-Agents/AdminOperations/src/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+  "function transferTokens(address token, address payable destination, uint256 amount)",
+  "function setImplementation(bytes4 selector, address implementation)",
+  "function registerPartner(address partner, uint256 _partnerShare, bool _noPositiveSlippage, bool _positiveSlippageToUser, uint16 _feePercent, string calldata partnerId, bytes calldata _data)",
+  "function setFeeWallet(address payable _feeWallet)",
+  "event RouterInitialized(address indexed router)",
+  "event AdapterInitialized(address indexed adapter)",
+];
+
+export const AUGUSTUS_SWAPPER_CONTRACT: string = "0xDEF171Fe48CF0115B1d80b88dc8eAB59176FEe57";
+export const SWAPPER_IFACE = new Interface(ADMIN_OPERATIONS);
```
Hyodar, line 14: Maybe this declaration could be in the `agent.spec.ts` file, since `SWAPPER_IFACE` is only used there and requires an `ethers` import?

Paraswap-Agents/AdminOperations/src/agent.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+import { Finding, FindingType, FindingSeverity, TransactionEvent, HandleTransaction } from "forta-agent";
+import { createAddress, TestTransactionEvent } from "forta-agent-tools/lib/tests";
+import { provideHandleTransaction } from "./agent";
+import { SWAPPER_IFACE } from "./utils";
+
+// AugustusSwapper address
+const SWAPPER_ADDR = createAddress("0xa1");
+// Address used to initiate transactions
+const USER_ADDR = createAddress("0xaa");
+// Irrelevent events/ functions used for tests
```
Hyodar, line 12: Really not a big deal, but there are some typos of "irrelevent", which should probably be "irrelevant".

Paraswap-Agents/AdminOperations/README.md
```diff
@@ -0,0 +1,63 @@
+# Admin Operations Agent
```
Kalzak, line 1: It would be nice if you had some examples transactions for improved testing (the transaction details for on-chain transactions are way more detailed than the ones in our test harnesses)
If it's not feasible to find transactions then it's fine, but I'd definitely say it's a nice-to-have.

Kalzak (approved):
Looks good!

Paraswap-Agents/AdminOperations/README.md
yasmine9: I definitely agree. I added the test transaction that I found and generates a finding.

PR #172, opened 2022-03-11
@paraswap/role changes

0xtaf:
# Role Changes

## Types of changes

- [x] New Agent
- [ ] Fix an agent error

## New Agent PR readiness

- [x] Agent contains tests

## Comments (optional)

### DIFF
--- a/Paraswap-Agents/Role-Changes/README.md
+++ b/Paraswap-Agents/Role-Changes/README.md
@@ -1,0 +1,29 @@
+# Role Change Detection Agent
+
+## Description
+
+This agent detects role changes in `AccessControl` contract.
+
+## Supported Chains
+
+- Ethereum
+- Polygon
+- BSC
+- Avalanche
+- Fantom
+- Ropsten
+
+## Alerts
+
+- PARASWAP-2
+  - Fired when `AccessControl` contract emits `RoleAdminChanged`, `RoleGranted`, `RoleRevoked` events.
+  - Severity is always set to "Info"
+  - Type is always set to "Info"
+  - Metadata fields when `RoleAdminChanged` event is emitted
+    - `role`: Role whose admin role is replaced.
+    - `previousAdminRole`: Previous admin role that is replaced with `newAdminRole`.
+    - `newAdminRole`: New admin role that is set as `role`'s admin role.
+  - Metadata fields when `RoleGranted` and `RoleRevoked` events are emitted
+    - `role`: Role that is granted or revoked.
+    - `account`: The account that is granted or revoked `role`.
+    - `sender`: The account that originated the contract call.
--- a/Paraswap-Agents/Role-Changes/README.md
+++ b/Paraswap-Agents/Role-Changes/README.md
@@ -2,7 +2,7 @@
 
 ## Description
 
-This agent detects role changes in `AccessControl` contract.
+This agent detects role changes in `AugustusSwapper` contract.
 
 ## Supported Chains
 
@@ -17,7 +17,7 @@ This agent detects role changes in `AccessControl` contract.
 
 - PARASWAP-2-1
 
-  - Fired when `AccessControl` contract emits `RoleAdminChanged` event.
+  - Fired when `AugustusSwapper` contract emits `RoleAdminChanged` event.
   - Severity is always set to "Info"
   - Type is always set to "Info"
   - Metadata fields:
@@ -27,7 +27,7 @@ This agent detects role changes in `AccessControl` contract.
 
 - PARASWAP-2-2
 
-  - Fired when `AccessControl` contract emits `RoleGranted` event.
+  - Fired when `AugustusSwapper` contract emits `RoleGranted` event.
   - Severity is always set to "Info"
   - Type is always set to "Info"
   - Metadata fields:
@@ -36,7 +36,7 @@ This agent detects role changes in `AccessControl` contract.
     - `sender`: The account that originated the contract call.
 
 - PARASWAP-2-3
-  - Fired when `AccessControl` contract emits `RoleRevoked` event.
+  - Fired when `AugustusSwapper` contract emits `RoleRevoked` event.
   - Severity is always set to "Info"
   - Type is always set to "Info"
   - Metadata fields:
--- a/Paraswap-Agents/Role-Changes/README.md
+++ b/Paraswap-Agents/Role-Changes/README.md
@@ -36,7 +36,7 @@ This agent detects role changes in `AccessControl` contract.
     - `sender`: The account that originated the contract call.
 
 - PARASWAP-2-3
-  - Fired when `AccessControl` contract emits `RoleGranted` event.
+  - Fired when `AccessControl` contract emits `RoleRevoked` event.
   - Severity is always set to "Info"
   - Type is always set to "Info"
   - Metadata fields:
--- a/Paraswap-Agents/Role-Changes/README.md
+++ b/Paraswap-Agents/Role-Changes/README.md
@@ -35,7 +35,7 @@ This agent detects role changes in `AccessControl` contract.
     - `account`: The account that is granted `role`.
     - `sender`: The account that originated the contract call.
 
-- PARASWAP-2-2
+- PARASWAP-2-3
   - Fired when `AccessControl` contract emits `RoleGranted` event.
   - Severity is always set to "Info"
   - Type is always set to "Info"
--- a/Paraswap-Agents/Role-Changes/README.md
+++ b/Paraswap-Agents/Role-Changes/README.md
@@ -8,7 +8,7 @@ This agent detects role changes in `AugustusSwapper` contract.
 
 - Ethereum
 - Polygon
-- BSC
+- Binance Smart Chain
 - Avalanche
 - Fantom
 - Ropsten
--- a/Paraswap-Agents/Role-Changes/README.md
+++ b/Paraswap-Agents/Role-Changes/README.md
@@ -15,15 +15,31 @@ This agent detects role changes in `AccessControl` contract.
 
 ## Alerts
 
-- PARASWAP-2
-  - Fired when `AccessControl` contract emits `RoleAdminChanged`, `RoleGranted`, `RoleRevoked` events.
+- PARASWAP-2-1
+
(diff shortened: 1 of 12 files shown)


Paraswap-Agents/Role-Changes/src/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+    metadata = {
+      role: log.args.role,
+      account: log.args.account,
+      sender: log.args.sender,
+    };
+  }
+  return Finding.fromObject({
+    name: `Role Change detected on AccessControl contract`,
+    description: `${log.name} event emitted`,
+    alertId: "PARASWAP-2",
```
RCantu92, line 29: Could have Findings such as `PARASWAP-2-1`, `PARASWAP-2-2`, and `PARASWAP-2-3` per each event. So that each event has a unique `alertId`.

Paraswap-Agents/Role-Changes/src/agent.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+      createAddress("0xa2"), // sender
+    ]);
+    // create a transaction with the previous event logs
+    const tx: TransactionEvent = new TestTransactionEvent().addAnonymousEventLog(CONTRACT, log.data, ...log.topics);
+    const findings: Finding[] = await handler(tx);
+
+    expect(findings).toStrictEqual([]);
+  });
+
+  it("should return single finding", async () => {
```
RCantu92, line 96: I think you could strengthen this test by adding a second anonymous event log to the transaction with `IrrelevantEvent` but confirming that it only returns a single correct one, the `RoleAdminChanged` event.

Paraswap-Agents/Role-Changes/src/agent.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+      formatBytes32String("ade"), // newAdminRole
+    ]);
+    // create a transaction with the previous event logs
+    const tx: TransactionEvent = new TestTransactionEvent().addAnonymousEventLog(CONTRACT, log1.data, ...log1.topics);
+    const findings: Finding[] = await handler(tx);
+
+    expect(findings).toStrictEqual([createFinding("RoleAdminChanged", log1.topics)]);
+  });
+
+  it("should return multiple findings", async () => {
```
RCantu92, line 110: Similar to the test before, I think this test could be strengthened by adding an event log with `IrrelevantEvent` as one of the middle logs (e.g. `log2` or `log3`) and confirming the agent "skip" the incorrect one and returns the three correct ones.

Also recommend not re-using test values for the event arguments.

RCantu92 (approved):

yasmine9 (approved):

Paraswap-Agents/Role-Changes/README.md
```diff
@@ -0,0 +1,45 @@
+# Role Change Detection Agent
+
+## Description
+
+This agent detects role changes in `AccessControl` contract.
```
Kalzak, line 5: The agent actually detects role changes in the `AugustusSwapper` contract. The contract `AccessControl` is defined, but it's actually inherited by `AugustusSwapper`.
I'd recommend changing all references to `AccessControl` to `AugustusSwapper` instead.

Paraswap-Agents/Role-Changes/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Supported Chains
+
+- Ethereum
+- Polygon
+- BSC
```
Kalzak, line 11: If all other chains are being referred to with their full name it's probably best to do the same for Binance Smart Chain too.

Paraswap-Agents/Role-Changes/src/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+  "event RoleAdminChanged(bytes32 indexed role, bytes32 indexed previousAdminRole, bytes32 indexed newAdminRole)",
+  "event RoleGranted(bytes32 indexed role, address indexed account, address indexed sender)",
+  "event RoleRevoked(bytes32 indexed role, address indexed account, address indexed sender)",
+];
+
+export const createFinding = (log: LogDescription) => {
+  if (log.name == "RoleAdminChanged") {
+    return Finding.fromObject({
+      name: "Admin role change detected on AccessControl contract",
+      description: `${log.name} event emitted`,
```
Kalzak, line 14: For all findings generated it seems more appropriate for the `name` to be swapped with `description`.
Typically the `name` is a short explanation with `description` going into a bit more detail.
For the findings in this agent it seems to be the opposite.
```suggestion
      name: `${log.name} event emitted`,
      description: "Admin role change detected on AccessControl contract",
```

Kalzak (commented):
formatting/readability recommendations

Paraswap-Agents/Role-Changes/src/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+export const EVENTS_ABI = [
+  "event RoleAdminChanged(bytes32 indexed role, bytes32 indexed previousAdminRole, bytes32 indexed newAdminRole)",
+  "event RoleGranted(bytes32 indexed role, address indexed account, address indexed sender)",
+  "event RoleRevoked(bytes32 indexed role, address indexed account, address indexed sender)",
+];
+
+export const createFinding = (log: LogDescription) => {
+  if (log.name == "RoleAdminChanged") {
+    return Finding.fromObject({
+      name: "Admin role change detected on AccessControl contract",
```
Kalzak, line 13: I recommend using a `switch` statement instead of an `if/else` chain here.
If you would like an example, see [here](https://github.com/NethermindEth/Forta-Agents/blob/%40paraswap/delegateCalls/Paraswap-Agents/delegate-calls/src/finding.ts)

Also, for the agents you build from now on I'd recommend trying to make one `createFinding` function that works for both your `agent.ts` and `agent.spec.ts`. 
It's going to be too much of an effort to make these changes to this existing agent, but when you're working on your next one keep it in mind.
The code for `createFinding` in the agent and tests are always very similar so it makes sense to try and have one function that works for both where possible.

Kalzak (changes requested):

Paraswap-Agents/Role-Changes/src/utils.ts
e1Ru1o: Not too much in line with this recommendation. 

Reasons:
- That function is going to be shared between the logic and the tests that can introduce errors. It is not something high probable to happen here because it is a pretty simple function and minimal sharing like this can be allowed, but in general sharing is not a good practice while testing.
- It is a lot easier to have different functions with custom parameters in testing an even many functions instead of only one `createFinding`, in order to achieve simplicity and clarity. For example I never used a `LogDescription` object in tests, so using the same function will add an extra work in the test to create that object

e1Ru1o: Also not worthy for a `request changes` comment because the approach used before works correctly, you are just suggesting a different approach with only syntax's changes

Paraswap-Agents/Role-Changes/src/agent.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+import { provideHandleTransaction } from "./agent";
+import { EVENTS_ABI } from "./utils";
+
+const TEST_SWAPPER = createAddress("0x1");
+const TEST_ACCESS_IFACE = new Interface(EVENTS_ABI);
+const IRRELEVANT_EVENT_IFACE = new Interface([
+  "event IrrelevantEvent(bytes32 indexed role, address indexed account, address indexed sender)",
+]);
+
+console.log(formatBytes32String("ROUTER_ROLE"));
```
e1Ru1o, line 13: This line should be removed

e1Ru1o (changes requested):

Paraswap-Agents/Role-Changes/src/agent.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+        newAdminRole: args[3],
+      };
+      break;
+    case "RoleGranted":
+      description = "Role grant detected on AugustusSwapper contract";
+      alertId = "PARASWAP-2-2";
+      metadata = {
+        role: args[1],
+        account: args[2].slice(0, 2) + args[2].slice(26),
+        sender: args[3].slice(0, 2) + args[3].slice(26),
```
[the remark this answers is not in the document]
Hyodar: Yeah, I agree with @e1Ru1o. Decoding the event log is actually really simple since we can use the `ethers.utils.Interface` object that was used to encode the event. For example:
```TS
const iface = new ethers.utils.Interface(["event EventName(address indexed addr)"]);

const encodedLog = iface.encodeEventLog(iface.getEvent("EventName"), [
  createAddress("0x1"), // addr
]);

const parsedLog = TEST_ACCESS_IFACE.parseLog(encodedLog);

console.log(parsedLog.args.addr)
```
By parsing the log inside the function or just sending the `parsedLog.args` to it as a parameter, there's no need to manually decode the addresses.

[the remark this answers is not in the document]
Hyodar: Actually, as I took a look at the `utils.ts` file, I realized that the `createFinding` function there is basically the testing `createFinding`, but instead of having `logName` and `args` as parameters, it has a `LogDescription` object, which is what you get through `parseLog`.

In this case, I think adapting the `createFinding` function to receive the parsed event log as a parameter isn't even needed since it could just be imported from `utils.ts`. Then, entries like `createFinding("RoleAdminChanged", log1.topics)` could be changed to `createFinding(TEST_ACCESS_IFACE.parseLog(log1))`. How do you feel about this, @e1Ru1o and @0xtaf ?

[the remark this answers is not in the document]
0xtaf: thank you @e1Ru1o @Hyodar for the great suggestions. I can do it the both ways, either with Interface's parseLog or decodeParameter. which one would you prefer here @e1Ru1o ?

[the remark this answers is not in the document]
Vxatz: The equivalent of `decodeParameter` would be `abiCoder.decode` in ethers.js, right?

PR #173, opened 2022-03-14
Sandwich-Detection

burraq101325:
# Sandwich

## Response

- [ ] New Agent
- [ ] Fix an agent error
- [x] Update an agent logic/test/documentation

### DIFF
--- a/sandwich/src/index.ts
+++ b/sandwich/src/index.ts
@@ -31,7 +31,7 @@ const factoryContract = new web3.eth.Contract(
 const usdcAddress = "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48"; // change
 const usdtAddress = "0xdac17f958d2ee523a2206206994597c13d831ec7"; // change
 const contractAddress = "0x3041CbD36888bECc7bbCBc0045E3B1f144466f5f";
-const Method_ID = ["0x8803dbee"]; //Method_ID of Uniswap Swaps "0x38ed1739" 
+const methodId = "0x8803dbee"; //methodId of Uniswap Swaps "0x8803dbee" 
 // const contractAddress = await factoryContract.methods
 // .getPair(usdcAddress, usdtAddress)
 // .call();
@@ -54,10 +54,10 @@ function provideHandleTransaction(
     // extract all swap events
     for (let i in txs) {
       const data = await web3.eth.getTransaction(txs[i]);
-      let ID = (data.input).slice(0,10);
+      const id = (data.input).slice(0,10);
 
       if (
-        Method_ID.includes(ID) &&
+        methodId === id &&
         data.to === contractAddress
       ) {
       const decodedData = abiDecoder.decodeMethod(data.input);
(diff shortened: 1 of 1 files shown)


sandwich/src/index.ts
```diff
@@ -26,11 +26,12 @@ abiDecoder.addABI(routerContract);
 const factoryContract = new web3.eth.Contract(
   factory as any,
   "0x5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6f"
-);
+);  
 
 const usdcAddress = "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48"; // change
 const usdtAddress = "0xdac17f958d2ee523a2206206994597c13d831ec7"; // change
 const contractAddress = "0x3041CbD36888bECc7bbCBc0045E3B1f144466f5f";
+const Method_ID = ["0x8803dbee"]; //Method_ID of Uniswap Swaps "0x38ed1739"
```
e1Ru1o, line 34: Why the comment is indicating a different ID?

e1Ru1o, line 34: Why the comment is indicating a different ID?

e1Ru1o: I would also suggest to use camel case for the name of this variable to keep uniformity

sandwich/src/index.ts
```diff
@@ -53,14 +54,13 @@ function provideHandleTransaction(
     // extract all swap events
     for (let i in txs) {
       const data = await web3.eth.getTransaction(txs[i]);
-
-      const decodedData = abiDecoder.decodeMethod(data.input);
+      let ID = (data.input).slice(0,10);
```
e1Ru1o, line 57: this var can be `const`

e1Ru1o, line 57: this var can be `const`

sandwich/src/index.ts
```diff
@@ -53,14 +54,13 @@ function provideHandleTransaction(
...
 
       if (
-        data.to === contractAddress &&
-        decodedData != undefined &&
-        decodedData.name === "swapTokensForExactTokens"
+        Method_ID.includes(ID) &&
```
e1Ru1o, line 60: I would prefer to here an equality check and get rid of the array because there is no need to use `array.includes` when the array contains only one element.

e1Ru1o, line 60: I would prefer to here an equality check and get rid of the array because there is no need to use `array.includes` when the array contains only one element.

sandwich/src/index.ts
burraq101325: I was doing the testing with two methods at first, I think I removed the wrong one. Sorry, will update it.
sure, will update the variable name.

sandwich/src/index.ts
burraq101325: sure, will update it.

sandwich/src/index.ts
burraq101325: I used the array.includes because I was looking for two method Ids (swapTokensForExactTokens 0x8803dbee and swapExactTokensForTokens 0x38ed1739). But then I found out when back running the victim transaction, the bots prefered the first one and it was also implemented in the logic. The includes or regular expression method can be used when multiple method Ids are to be considered, but here it's not needed will remove the array and use the equality check.

PR #176, opened 2022-03-18
Delegate votes

Hyodar:
# Delegate votes

## Types of changes

- [x] New Agent
- [ ] Fix an agent error

## New Agent PR readiness

- [x] Agent contains tests
- [x] Agent can recover from failed error calls when they can fail (e.g. errors are being caught on calls that can fail)


Benqi-Agents/Large-Delegate-Votes-Change/src/agent.ts
```diff
@@ -0,0 +1,55 @@
...
+  const description = `There was a >= ${percentageThreshold.toString()}% increase in delegate votes based on the previous amount`;
+
+  return async (txEvent: TransactionEvent): Promise<Finding[]> => {
+    const findings: Finding[] = [];
+
+    txEvent.filterLog(DELEGATE_VOTES_CHANGED_ABI, contractAddress).forEach((log: LogDescription) => {
+      if (!isLarge(log.args.previousBalance, log.args.newBalance, percentageThreshold)) {
+        return;
+      }
+
```
yasmine9, line 30: I think this return can be avoided by changing the if statement condition to `if (isLarge.. ) findings.push(...) `

Benqi-Agents/Large-Delegate-Votes-Change/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Supported Chains
+
+- Avalanche
+
+## Alerts
+
+- BENQI-1-1
+  - Fired when a delegate account's vote balance increases by 30% or more.
+  - Severity is always set to "Info"
```
yasmine9, line 17: The agent takes the percentage as a parameter, so avoid using `30%` in describing the agent and the finding, referring to this as a large increase should be enough.

Benqi-Agents/Large-Delegate-Votes-Change/README.md
```diff
@@ -0,0 +1,37 @@
...
+vote balance by monitoring `DelegateVotesChanged` events emitted by the token
+contract.
+
+## Supported Chains
+
+- Avalanche
+
+## Alerts
+
+- BENQI-1-1
```
yasmine9, line 15: Since we have only one finding, there's no need to have the sub-number in the alert Id, `BENQI-1` is enough imo

Benqi-Agents/Large-Delegate-Votes-Change/src/agent.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+      txEvent.addAnonymousEventLog(QI_ADDRESS, alertLog2.data, ...alertLog2.topics);
+
+      const findings = await handleTransaction(txEvent);
+
+      expect(findings).toStrictEqual([
+        createFinding(DELEGATE_ADDRESS, "100", increaseByPercentage("100", PERCENTAGE_THRESHOLD)),
+        createFinding(DELEGATE_ADDRESS, "200", increaseByPercentage("200", PERCENTAGE_THRESHOLD)),
+      ]);
+    });
+  });
```
yasmine9, line 121: To avoid having all the positive test cases with an increase always equal to the threshold, You can make some with a greater value (`PERCENTAGE_THRESHOLD + 10` for example)

Benqi-Agents/Large-Delegate-Votes-Change/README.md
```diff
@@ -0,0 +1,37 @@
...
+  - Severity is always set to "Info"
+  - Type is always set to "Info"
+  - Metadata contains:
+    - `delegate`: Address of the related delegate
+    - `previousBalance`: Previous delegate vote balance
+    - `newBalance`: Current delegate vote balance
+
+## Test Data
+
+The agent behaviour can be verified with the following transactions:
```
yasmine9, line 26: If all the txs below generate findings, it would be better to mention that. I suggest ,if possible, adding some negative cases where no finding is returned.

yasmine9 (changes requested):

Benqi-Agents/Large-Delegate-Votes-Change/src/agent.ts
Hyodar: Sure, I did it this way because I thought readability was better than having one more level of nesting, but it's honestly not that big of a change in this case, so I'm changing this.

Benqi-Agents/Large-Delegate-Votes-Change/README.md
Hyodar: I removed some positive cases (they keep being positive, I just thought it was too much cases) and added some negative ones.

e1Ru1o (changes requested):
- `npm install` show one high severity -->` npm audit fix` required

e1Ru1o (approved):
No need to add a PoC, because the agent is very simple and have enough test data in the README

Benqi-Agents/Large-Delegate-Votes-Change/README.md
```diff
@@ -0,0 +1,33 @@
...
+balance by monitoring `DelegateVotesChanged` events emitted by the token
+contract.
+
+## Supported Chains
+
+- Avalanche
+
+## Alerts
+
+- BENQI-1-1
```
yasmine9, line 15: Please update the alert Id in the README as well

Hyodar: Sure, this passed me by.

Hyodar: Fixed in 0e0b3b0, described how to change the threshold as well.

yasmine9 (approved):

PR #180, opened 2022-03-21
@benqi/Large-Withdrawals-From-Pools

Kalzak:
# Large Withdrawals From Pools

## Types of changes

- [x] New Agent
- [ ] Fix an agent error

## New Agent PR readiness

- [x] Agent contains tests
- [x] Agent can recover from failed error calls when they can fail (e.g. errors are being caught on calls that can fail)

## Comments (optional)


Benqi-Agents/Large-Pool-Withdrawal/src/agent.ts
```diff
@@ -0,0 +1,100 @@
...
+  // Create an ethers contract instance to query on-chain data
+  const qiTokenContract = new ethers.Contract(
+    qiToken,
+    QITOKEN_IFACE.format(ethers.utils.FormatTypes.full),
+    getEthersProvider()
+  );
+  try {
+    return await qiTokenContract.totalSupply({ blockTag: blockNumber });
+  } catch (error) {
+    // Not sure what to do when there's an error
```
Kalzak, line 33: Thoughts on what to do when there is an error in the Web3 call?
Should we assume it's within the threshold or outside the threshold?

Benqi-Agents/Large-Pool-Withdrawal/src/agent.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    // Convert the QiToken set into an array
+    const qiTokens: string[] = Array.from(qiTokensSet);
+
+    // For each QiToken
+    await Promise.all(
+      qiTokens.map(async (qiToken) => {
+        // Get all `Redeem` events from QiTokens
+        const redeems = tx.filterLog(QITOKEN_IFACE.getEvent("Redeem").format("full"), qiToken);
+        // For each `Redeem` event
```
yasmine9, line 76: Instead of looping over `qiTokens`, you can use `FilterLog` once and give it the `qiTokens` array as an argument, the loop is implemented inside that function.

yasmine9 (changes requested):

Benqi-Agents/Large-Pool-Withdrawal/src/agent.ts
yasmine9: Not sure about this, it would make more sense to consider it below the threshold, to avoid giving false alerts. 
But I believe it's important to point out that the agent is running into an issue, otherwise it will never give alerts. 
Maybe returning an `error Finding` in that case (and for all our agents) is a good idea, the finding will include a description of the problem encountered (ex. `Agent unable to fetch the totalSupply from qiToken contract`).

Benqi-Agents/Large-Pool-Withdrawal/README.md
```diff
@@ -0,0 +1,28 @@
...
+    - `qiToken`: The address of the pool where the withdrawal occurred
+    - `totalSupply`: The total supply before the transaction (supply snapshot taken at previous block)
+    - `redeemTokens`: The amount of tokens retrieved in the transaction
+
+## Test Data
+
+The agent behaviour can be verified with the following transactions:
+
+- 0x92b971e2c84b79cf7df949ee3645f0e28310ac0fe94b7d436f34236dc8fc8d14 
+  - (Withdrawal of `7264199798` from total supply of `35508760798710315`)
```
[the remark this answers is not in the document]
Kalzak: It's such a small percentage that it's basically [illegible](https://www.wolframalpha.com/input?i=7264199798%2F35508760798710315) and wouldn't look great in the documentation, which is why I decided to write it as "withdrawal of `x` from total supply of `y`"

0xtaf:
.gitignore, .prettierrc and .dockerignore files are missing.

e1Ru1o (changes requested):
PoC missing

Benqi-Agents/Large-Pool-Withdrawal/src/agent.ts
```diff
@@ -0,0 +1,100 @@
...
+        await Promise.all(
+          redeems.map(async (redeem) => {
+            // Get the number of tokens redeemed
+            const redeemTokens = redeem.args.redeemTokens;
+            // Get the total supply from the previous block
+            const totalSupply = await getTotalSupply(qiToken, tx.blockNumber - 1);
+            // If the amount of tokens redeemed is more than `thresholdPercentage` of `totalSupply`
+            if (totalSupply.div(redeemTokens).lt(1 / thresholdPercentage)) {
```
Hyodar, line 84: There's something I've noticed when working on my agent (also dealing with threshold percentages) that I think would be good to comment here.

One of the best ways I found for doing this kind of comparison is this one:
```TS
// `thresholdPercentage`, in this case, is a BigNumber like "30" for 30%
redeemTokens.gte(totalSupply.mul(thresholdPercentage).div(100));
```
This avoids some problems, mainly divisions by zero, and sticks with `BigNumber`s. In this case, `thresholdPercentage` can be zero, as well as `redeemTokens`, and there won't be any errors. The only problem with this approach is that the threshold needs to be a percentage integer.

Benqi-Agents/Large-Pool-Withdrawal/src/agent.ts
0xtaf: why do we need to 'fetch' total supply in the first place? isn't it a specific number that will never change?

Kalzak: The naming structure is quite confusing but these qiTokens are different to QI tokens.
See [qiToken](https://docs.benqi.fi/resources/contracts/benqi-liquidity-markets) addresses vs [QI Token](https://docs.benqi.fi/resources/contracts/benqi-liquidity-markets) address.

Kalzak: As for considering failures, after a discussion with Lazaro I will be removing the `try/catch` from the code as I have only written it to protect against network issues rather than any potential wrong calls.

Benqi-Agents/Large-Pool-Withdrawal/README.md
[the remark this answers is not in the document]
Kalzak: I have found a much better transaction to demonstrate the agents capabilities and have updated the README.

Benqi-Agents/Large-Pool-Withdrawal/src/abi.ts
```diff
@@ -0,0 +1,1648 @@
+import { ethers } from "ethers";
+
+const COMPTROLLER_ABI = [
+  { inputs: [], payable: false, stateMutability: "nonpayable", type: "constructor" },
+  {
```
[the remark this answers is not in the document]
Kalzak: ABI has been replaced with specific functions and events in string format.

Benqi-Agents/Large-Pool-Withdrawal/src/agent.ts
```diff
@@ -0,0 +1,97 @@
...
+export const getTotalSupply = async (qiToken: string, blockNumber: any) => {
+  // Create an ethers contract instance to query on-chain data
+  const qiTokenContract = new ethers.Contract(
+    qiToken,
+    QITOKEN_IFACE.format(ethers.utils.FormatTypes.full),
+    getEthersProvider()
+  );
+
+  return await qiTokenContract.totalSupply({ blockTag: blockNumber });
+}
```
Hyodar, line 32: It seems we don't need a provider call since the QI total supply is constant, see [here](https://snowtrace.io/address/0x8729438eb15e2c8b576fcc6aecda6a148776c0f5#code).
I was also querying it on-chain before I took a look at the contract and noticed that.

Benqi-Agents/Large-Pool-Withdrawal/src/agent.ts
Kalzak: This has been addressed in the following commit.

Benqi-Agents/Large-Pool-Withdrawal/src/agent.ts
```diff
@@ -0,0 +1,100 @@
...
+    COMPTROLLER_ADDR,
+    [COMPTROLLER_IFACE.getFunction("getAllMarkets").format("full")],
+    getEthersProvider()
+  );
+  // Get all QiTokens
+  const qiTokens = await comptrollerContract.getAllMarkets();
+  // Add each qiToken to the set `QITOKENS`
+  qiTokens.forEach((qiToken: string) => {
+    QITOKENS.add(qiToken.toLowerCase());
+  });
```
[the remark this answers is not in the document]
Kalzak: Sets have been replaced by arrays in the following commit.

Benqi-Agents/Large-Pool-Withdrawal/src/agent.ts
Kalzak: https://github.com/NethermindEth/Forta-Agents/pull/180#discussion_r833120248

Benqi-Agents/Large-Pool-Withdrawal/src/agent.ts
Kalzak: Nice! That's a much better way to approach percentages!

Benqi-Agents/Large-Pool-Withdrawal/src/agent.ts
Hyodar: You're right!

Benqi-Agents/Large-Pool-Withdrawal/src/agent.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+const QITOKEN_ADDR = createAddress("0xa1");
+const NEW_QITOKEN_ADDR = createAddress("0xa2");
+const COMPTROLLER_ADDR = createAddress("0xb1");
+const EXTERNAL_ADDR = createAddress("0xc1");
+const USER_ADDR = createAddress("0xd1");
+
+describe("Benqi Finance Large Pool Withdrawal Agent Test Suite", () => {
+  let handler: HandleTransaction;
+  let mockGetTotalSupply: any;
+  //let initializeSet: Set<string>;
```
yasmine9, line 23:
```suggestion
```

yasmine9 (changes requested):

Benqi-Agents/Large-Pool-Withdrawal/README.md
```diff
@@ -0,0 +1,30 @@
...
+  - Metadata:
+    - `qiToken`: The address of the pool where the withdrawal occurred
+    - `totalSupply`: The total supply before the transaction (supply snapshot taken at previous block)
+    - `redeemTokens`: The amount of tokens retrieved in the transaction
+
+## Test Data
+
+The agent behaviour can be verified with the following transactions:
+
+- 0xb7b8e8bddfe9b0fbef66abb9cc8d316c6c63c5a920f47fa0d5a3dab234b2c7fb (Multiple withdrawals)
```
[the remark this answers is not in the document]
e1Ru1o: It doesn't work for me either, 0 findings in all the cases

yasmine9 (approved):

yasmine9 (commented):
- Please change `contracts` folder name to `PoC`

- `npm install` showing 1 high severity vulnerability, please run `npm audit fix`

e1Ru1o (approved):

yasmine9 (approved):

PR #201, opened 2022-04-26
DD15 - Insolvent Contract

RCantu92:
# Insolvent Contract Bot

## Types of changes

- [x] New Agent
- [ ] Fix an agent error

## New Agent PR readiness

- [x] Agent contains tests
- [ ] Agent can recover from failed error calls when they can fail (e.g. errors are being caught on calls that can fail)

## Comments (optional)


e1Ru1o (changes requested):

DYDX-Bots/Insolvent_Contract/src/balance.fetcher.ts
```diff
@@ -1,50 +1,37 @@
...
-      )
-    );
+    const balance = await this.proxyContract.getTotalBorrowerDebtBalance({ blockTag: block });
     this.cache.set(key, balance);
     return balance;
   }
 
   public async getTotalActiveBalanceCurrentEpoch(block: string | number): Promise<BigNumber> {
-    const key: string = `totalActiveBalanceCurrentEpoch - ${block}`;
+    const key: string = `${this.proxyContract} - totalActiveBalanceCurrentEpoch - ${block}`;
```
e1Ru1o, line 31: `this.proxyContract` is not relevant in the key

DYDX-Bots/Insolvent_Contract/README.md
```diff
@@ -0,0 +1,31 @@
+# Insolvent Contract Bot
+
+## Description
+
+This bot detects when the contract is insolvent. Insolvency is refers to when `totalBorrowerDebtBalance` is greater than `totalActiveBalanceCurrentEpoch`. The checks for both variables in every block.
+
```
yasmine9, line 6: ` The checks for both variables in every block.` Do you mean `the bot checks ..`?

DYDX-Bots/Insolvent_Contract/README.md
```diff
@@ -0,0 +1,31 @@
...
+> Note: To test the following, follow the comments in `utils.ts`.
+
+The bot behaviour can be verified with the following contracts on the Kovan ETH testnet:
+
+[0x4aaAF7a1a829b2130AF2Ec7A4A4F5FEd27D8eD1F](https://kovan.etherscan.io/address/0x4aaAF7a1a829b2130AF2Ec7A4A4F5FEd27D8eD1F) - `TestImplemenation`.
+
+[0xe9511Faa2B2ccE548A5999b4bC3772e6a0f1C14A](https://kovan.etherscan.io/address/0xe9511Faa2B2ccE548A5999b4bC3772e6a0f1C14A) - `TestProxy`.
+
+You can test the bot on the Kovan ETH testnet by running `npm start`. Also, you can use the setter functions in the `testProxy` contract to toggle both `totalBorrowerDebtBalance` and `totalActiveBalanceCurrentEpoch` to test them in real time. This is because the bot checks these values in every block.
```
yasmine9, line 31: I'm not sure why this part is needed in the note: `This is because the bot checks these values in every block.`

DYDX-Bots/Insolvent_Contract/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "dydx-15",
+  "version": "0.0.1",
+  "description": "Detect insolvent contract",
```
yasmine9, line 4:
```suggestion
  "description": "Detect insolvent contract in dYdX Liquidity Module",
```
a suggestion to improve description

DYDX-Bots/Insolvent_Contract/src/findings.ts
```diff
@@ -0,0 +1,17 @@
+import { Finding, FindingType, FindingSeverity } from "forta-agent";
+import { BigNumber } from "ethers";
+
+export const createFinding = (totalBorrowerDebtBalance: BigNumber, totalActiveBalanceCurrentEpoch: BigNumber) => {
+  return Finding.fromObject({
+    name: "Contract Insolvent",
```
yasmine9, line 6:
```suggestion
    name: "Liquidity Module Contract is insolvent",
```
It would be better to include what contract is insolvent.

DYDX-Bots/Insolvent_Contract/src/findings.ts
```diff
@@ -0,0 +1,17 @@
...
+    description: "Total borrower debt balance has exceeded total active balance current epoch",
```
yasmine9, line 7:
```suggestion
    description: "The total borrowed balance has exceeded total active balance in the current epoch",
```
The current one is fine, this is a suggestion to make the description more inspired from the docs rather than the variables/events names. Using "Total borrowed balance" seems more understandable than Total borrower debt balance.

DYDX-Bots/Insolvent_Contract/src/balance.fetcher.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+      expect(fetchedDebtBalance).toStrictEqual(totalBorrowerDebtBalance);
+    }
+
+    // clear mock to use cache
+    mockProvider.clear();
+    for (let [totalBorrowerDebtBalance, , blockNumber] of TEST_CASES) {
+      createGetTotalBorrowerDebtBalance(totalBorrowerDebtBalance, blockNumber);
+
+      const fetchedDebtBalance: BigNumber = await fetcher.getTotalBorrowerDebtBalance(blockNumber);
+      expect(fetchedDebtBalance).toStrictEqual(totalBorrowerDebtBalance);
```
yasmine9, line 55: In order to use cache, we don't need to call `createGetTotalBorrowerDebtBalance` again. Same note applies to the following test case.

DYDX-Bots/Insolvent_Contract/src/agent.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  BigNumber.from("20000000000000000000"), // 20
+];
+const testActiveBalanceCurrentEpoch: BigNumber[] = [
+  BigNumber.from("1000000000000000000"), // 1
+  BigNumber.from("9000000000000000000"), // 9
+  BigNumber.from("15000000000000000000"), // 15
+  BigNumber.from("22000000000000000000"), // 22
+];
+
+describe("high tether transfer agent", () => {
```
yasmine9, line 24: Please update description

DYDX-Bots/Insolvent_Contract/src/agent.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+      inputs: [],
+      outputs: [totalActiveBalanceCurrentEpoch],
+    });
+  };
+
+  const handleBlock: HandleBlock = provideHandleBlock(balanceFetcher, testThreshold);
+
+  beforeEach(() => mockProvider.clear());
+
+  it("should detect a contract in an insolvent state", async () => {
```
yasmine9, line 46: I suggest adding a test case where the bot ignores an insolvency case in a different contract

DYDX-Bots/Insolvent_Contract/README.md
yasmine9: I think you can generate a `test block` where there's an insolvency case by running these setters.

yasmine9 (changes requested):

DYDX-Bots/Insolvent_Contract/src/agent.spec.ts
RCantu92: Thinking about, I'm not sure how helpful this will be. Since this isn't `handleTransaction` that detects where an event is emitted, but instead proactively calls a specific contract's function at a specific address that was passed as an argument when the `balanceFetcher` was created. This bot is more proactive, than reactive.

Then because the contract's address was passed as an argument, it wouldn't call the wrong contract.

DYDX-Bots/Insolvent_Contract/README.md
RCantu92: > I'm not sure why this part is needed in the note: This is because the bot checks these values in every block.

It was me trying to explain why you could simply run `npm start` and toggle the values in real time.

> I think you can generate a test block where there's an insolvency case by running these setters.

I tried that several times, but what kept happening, is that at some point the block would be "too old" (this is my assumption) and it wouldn't generate a finding anymore and would actually error out.  I thought may it had to do with the provided RPC endpoint, since the bot would work again if I simply toggled the values again.

DYDX-Bots/Insolvent_Contract/README.md
```diff
@@ -0,0 +1,32 @@
+# Insolvent Contract Bot
+
+## Description
+
+This bot detects when the contract is insolvent. Insolvency refers to when the difference between `totalBorrowerDebtBalance` and `totalActiveBalanceCurrentEpoch` is greater than a set static threshold. The bot checks for both variables in every block.
+
+> Note: To set the static threshold to a desired amount, update the value of `THRESHOLD_AMOUNT` in `utils.ts`. Currently, it is set to `0`, and thus will detect **any** difference between the two varibles.
```
yasmine9, line 7:
```suggestion
> Note: To set the static threshold to a desired amount, update the value of `THRESHOLD_AMOUNT` in `utils.ts`. Currently, it is set to `0`, and thus will detect **all** insolvency cases.
```
I suggest this to help the client understand this direct link between this threshold and the insolvency detection.

DYDX-Bots/Insolvent_Contract/src/balance.fetcher.ts
```diff
@@ -0,0 +1,45 @@
...
+    const key: string = `${this.moduleContract} - totalBorrowerDebtBalance - ${block}`;
+    if (this.cache.has(key)) return this.cache.get(key) as Promise<BigNumber>;
+
+    const balance = await this.moduleContract.getTotalBorrowerDebtBalance({ blockTag: block });
+    this.cache.set(key, balance);
+    return balance;
+  }
+
+  public async getTotalActiveBalanceCurrentEpoch(block: string | number): Promise<BigNumber> {
+    const key: string = `${this.moduleContract} - totalActiveBalanceCurrentEpoch - ${block}`;
```
yasmine9, line 38: I think we can remove `${this.moduleContract}` from the keys, it's the address of the contract being called so it can never be different.

DYDX-Bots/Insolvent_Contract/src/agent.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+];
+const testActiveBalanceCurrentEpoch: BigNumber[] = [
+  BigNumber.from("1000000000000000000"), // 1
+  BigNumber.from("9000000000000000000"), // 9
+  BigNumber.from("15000000000000000000"), // 15
+  BigNumber.from("22000000000000000000"), // 22
+];
+
+describe("Insolvent Contract test suite", () => {
+  const mockProvider: MockEthersProvider = new MockEthersProvider();
```
yasmine9, line 25: I suggest adding a test case where different blocks are being handled by the bot and multiple findings are generated, since all test cases are with a single block.

yasmine9 (commented):
I left some suggestions, mainly adding a test case for the bot where multiple blocks are handled. 
PR will be approved after the discussion about test blocks is resolved.

yasmine9 (approved):

e1Ru1o (approved):

PR #210, opened 2022-05-02
DD16 - Large Borrower Allocation

RCantu92:
# Large Borrower Allocation

## Types of changes

- [x] New Agent
- [ ] Fix an agent error

## New Agent PR readiness

- [x] Agent contains tests
- [ ] Agent can recover from failed error calls when they can fail (e.g. errors are being caught on calls that can fail)

## Comments (optional)


e1Ru1o (approved):
Please add automatic network data selection using initialize function

DYDX-Bots/Large_Borrower_Allocation/src/findings.ts
```diff
@@ -0,0 +1,19 @@
+import { LogDescription, Finding, FindingSeverity, FindingType } from "forta-agent";
+
+export const createFinding = (log: LogDescription) => {
+  return Finding.fromObject({
+    name: "Large borrower allocation change detected on dYdX perpetual exchange.",
```
yasmine9, line 5: I think using `change` can be understood as a large difference between the old and new allocation, while this bot is detecting borrowers with large allocation.

DYDX-Bots/Large_Borrower_Allocation/src/agent.ts
```diff
@@ -0,0 +1,33 @@
...
+import { createFinding } from "./findings";
+
+const networkManager = new NetworkManager();
+
+export const provideInitialize = (provider: providers.Provider) => async () => {
+  const { chainId } = await provider.getNetwork();
+  networkManager.setNetwork(chainId);
+};
+
+export function provideHandleTransaction(networkManager: NetworkData, threshold: BigNumber): HandleTransaction {
```
yasmine9, line 15: Any specific reason why the networkManager parameter has a `NetworkData` type not a `NetworkManager`?

yasmine9 (approved):
One minor comment and a question concerning the networkManager.
The bot logic and tests LGTM ✅

DYDX-Bots/Large_Borrower_Allocation/src/agent.ts
RCantu92: When I was going off the ApeSwap bot that first implemented it, the type was set as `NetworkData`, and me trying to make sure I got it to work properly, use the same. However, I just switched it to `NetworkManager` to see if it would break, and it still functions as expected. Perhaps they can be interchangeable? Might make sense since the `NetworkManager` class `implements` the `NetworkData` interface.

e1Ru1o: `NetworkData` is correct, in fact the idea is to use a `NetworkData` you can even change the name of the variable, because the int terms of the bot logic it only needs a NetworkData, because it is not going to Manage the network anyway.

yasmine9 (commented):
Approving after adding the error handle to the network manager.

yasmine9 (approved):

PR #230, opened 2022-05-19
UM09 - Lending pool reentrancy

yehia67:
# Lending pool Reentrancy Detector

## Types of changes

- [x] New Bot
- [ ] Fix a bot error

## New Bot PR readiness

- [x] Bot contains tests


Umee-Bots/lendingPool-reentrancy/src/agent.ts
```diff
@@ -0,0 +1,38 @@
...
+    const { traces } = txEvent;
+    for (let i = 0; i < traces.length; i++) {
+      const trace = traces[i];
+      const depth = trace.traceAddress.length;
+
+      if (trace.action.to === contractAddress) {
+        let j;
+
+        for (j = i + 1; j < traces.length; j++) {
+          if (traces[j].traceAddress.length === depth) {
```
Hyodar, line 18: Actually, in this case it'd be `traces[j].traceAddress.length <= depth`, right?
Example: `[[], [0], [0, 0], [0, 0, 0], [1]]`. Let's say `[0, 0, 0]` is a call to `LendingPool`, then the subtree ends in `[1]`.

Umee-Bots/lendingPool-reentrancy/src/agent.ts
```diff
@@ -0,0 +1,38 @@
...
+            i = j;
```
Hyodar, line 19: Also, this should be `i = j - 1` right? Otherwise the element at index `j` will be ignored.

Umee-Bots/lendingPool-reentrancy/src/agent.ts
```diff
@@ -0,0 +1,38 @@
...
+            break; // subtree ended
+          }
+          if (traces[j].action.to === contractAddress) {
+            reenterancyFunctionsSignatures.map((functionSignature) => {
+              try {
+                utils.FUNCTIONS_INTERFACE.decodeFunctionData(functionSignature.split("(")[0], txEvent.transaction.data);
+                findings.push(utils.createFinding(functionSignature.split("(")[0]));
+              } catch (e) {}
+            });
```
Hyodar, line 28: This almost works, but you're examining the transaction (not the trace) data to get the signature.
Here's my suggestion:
- Create `REENTRANCY_FUNCTIONS_SELECTORS`, like this (also fix the typo "reenterancy" while you're at it):
```TS
// utils.ts
const REENTRANCY_FUNCTIONS_SELECTORS = Object.values(FUNCTIONS_INTERFACE.functions).map(el => iface.getSighash(el));
```
- In this part I selected, instead do:
```TS
const selector = (traces[j].action.input || "").slice(0, 10); // "0x" and first 4 bytes
if (REENTRANCY_FUNCTIONS_SELECTORS.includes(selector)) {
  // push finding
  break; // no need to check other traces, we just need to know if one case happened
}
```

Umee-Bots/lendingPool-reentrancy/src/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+    name: "Lending Pool Reentrancy",
+    description: `${transactionName} transaction is reentrant`,
+    alertId: "Um-09",
+    type: FindingType.Exploit,
+    severity: FindingSeverity.High,
+    metadata: {
+      transactionName: transactionName,
+    },
+  });
+};
```
Hyodar, line 31: It seems like this is considering a function name is the "transaction name", which may be a bit confusing. My suggestion would be that you just say, in name and description, that a reentrancy happened in the LendingPool contract. As for the metadata, I think it'd be interesting to pass the selector of the initial call to `LendingPool` (the one that, when we find, we check the call subtree), and the selector of the first reentrancy case we find.

Umee-Bots/lendingPool-reentrancy/src/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+  "function flashLoan(address receiverAddress,address[] calldata assets,uint256[] calldata amounts,uint256[] calldata modes,address onBehalfOf,bytes calldata params,uint16 referralCode)",
+];
+
+const FUNCTIONS_INTERFACE = new Interface(REENTERANCY_FUNCTIONS_ABI);
+
+const createFinding = (transactionName: string): Finding => {
+  return Finding.fromObject({
+    name: "Lending Pool Reentrancy",
+    description: `${transactionName} transaction is reentrant`,
+    alertId: "Um-09",
```
Hyodar, line 24: Also, it's probably better to use `UMEE-9` as `alertId` so it's like other bots we did

Umee-Bots/lendingPool-reentrancy/README.md
```diff
@@ -0,0 +1,18 @@
+# Detect Reentrancy In Lending Pool
+
+## Description
+
+Detects reentrancy in any call to LendingProtocol. If that’s the case, emits an alert.
```
Hyodar, line 5: This was changed, so it's better to make it clear in the README. Now we only emit an alert if the reentrant call function is in a blacklist.

Umee-Bots/lendingPool-reentrancy/src/agent.ts
```diff
@@ -0,0 +1,40 @@
...
+    for (let i = 0; i < traces.length; i++) {
+      const trace = traces[i];
+      const depth = trace.traceAddress.length;
+      if (trace.action.to === config.lendingPoolAddress) {
+        let j;
+
+        for (j = i + 1; j < traces.length; j++) {
+          if (traces[j].traceAddress.length <= depth) {
+            i = j - 1;
+            continue; // subtree ended, non reentrant call
```
Hyodar, line 21: Can you explain this change?

Umee-Bots/lendingPool-reentrancy/src/agent.ts
```diff
@@ -0,0 +1,40 @@
+import { Finding, HandleTransaction, TransactionEvent } from "forta-agent";
+
+import CONFIG from "./agent.config";
+
+import utils, { AgentConfig } from "./utils";
+
+export const handleTransaction =
```
Hyodar, line 7: This should probably be `provideHandleTransaction` instead.

Umee-Bots/lendingPool-reentrancy/src/agent.config.ts
```diff
@@ -0,0 +1,12 @@
+import utils, { AgentConfig } from "./utils";
+
+const CONFIG: AgentConfig = {
+  reentrancyFunctionsSelectors: utils.REENTRANCY_FUNCTIONS_SELECTORS,
```
Hyodar, line 4: Maybe you could make a `reentrancyBlacklist` field instead (I think it's more descriptive than `reentrancyFunctions`) so that the client can input a list of signatures, like this:
```
{
  reentrancyBlacklist: [
    "signature()",
    // [...]
  ],
  // [...]
}
```

Then, inside `provideHandleTransaction`, you could compute the sighashes, like this:
```TS
export const handleTransaction = (config: AgentConfig): HandleTransaction => {
  const sighashes = getSighashes(config.reentrancyBlacklist);
  return async (txEvent: TransactionEvent): Finding[] => {
    // [...]
  }
```

Hyodar: This would make it better, at least in a way, because the configuration would be centralized in `agent.config.ts`. If you do make these changes, you can also remove `REENTRANCY_FUNCTIONS_SIGNATURES` and `REENTRANCY_FUNCTIONS_ABI`, since they won't be used anymore.
Also, `getSighashes` would do something similar to what's being done currently - get `reentrancyBlacklist`, create an Interface from it and get the sighashes of all the functions.

Umee-Bots/lendingPool-reentrancy/src/agent.ts
yehia67: If I have multiple external calls on the same level. The first one is not reentrant, but the second one is. It will `break;` before checking it.
For example, on the PoC inside the `fallback` function, if I added any function before the `deposit,` it will not detect the reentrant call.

Hyodar: I think `break` is really what's intended. If we find a call to `LendingPool`, we go through that call's subtree and try finding another call to `LendingPool` which has a blacklisted selector. When the subtree ends, we can't consider anymore that the call that originated the current one is a call to `LendingPool`, because it's not in the subtree.

Actually, your PoC shouldn't emit any findings, because there's no reentrancy in the `LendingPool` contract, there's reentrancy in your contract!

Hyodar: For example, there's no `target contract` reentrancy here:
```
some contract
+- target contract
+- some contract
  +- target contract
```
In this case, there is:
```
some contract
+- target contract
  +- target contract
```

Umee-Bots/lendingPool-reentrancy/src/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const getSigHashes = (signaturesArray: string[]): string[] => {
+  return signaturesArray.map((signature) => {
+    return encodeFunctionSignature(signature);
+  });
+};
+
+const createFinding = (initialCallSelector: string, lendingPoolCallSelector: string): Finding => {
+  return Finding.fromObject({
+    name: "Detect Lending Pool Reentrancy",
```
Hyodar, line 17: Suggestion: something like `"LendingPool call reentrancy"`.

Umee-Bots/lendingPool-reentrancy/src/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+    description: "Detect when a reentrancy happened in the LendingPool contract",
```
Hyodar, line 18: Suggestion: something like `"Reentrancy detected in call to LendingPool contract"`

Hyodar:
Now, commenting about the PoC. In this case, your PoC does work, but it's not obvious why it is like that and it could give the wrong idea that we're detecting reentrancy in the caller contract - thus, simpler could be a lot better. 
My suggestion: you can make just a simple contract that has some of the signatures `LendingPool` has, and just make one function reentrant. For example:
```Solidity
pragma solidity 0.6.12;

contract MockLendingPool {
    function deposit(
        address asset,
        uint256 amount,
        address onBehalfOf,
        uint16 referralCode
    ) public payable {
        // making a reentrant call to withdraw()
        MockLendingPool(address(this)).withdraw(address(0x0), 0, address(0x0));
    }

    function withdraw(
        address asset,
        uint256 amount,
        address to
    ) public payable {
        // no-op
    }

    function test() public {
        deposit(address(0x0), 0, address(0x0), 0);
    }
}
```
If you make the bot process a transaction with a call to `test()`, it should emit a finding.

Hyodar (changes requested):
Requesting the changes I mentioned.

Umee-Bots/lendingPool-reentrancy/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "umee-9",
+  "version": "0.0.1",
+  "description": "Detects reentrancy in reentrant calls inside LendingProtocol smart contract. If that’s the case, emits an alert.",
+  "chainIds": [1],
+  "repository": "https://github.com/NethermindEth/Forta-Agents/tree/umee/UM09-lendingpool-Reentrancy",
```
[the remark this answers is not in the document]
yehia67: renaming the branch will close the PR😢 but I changed the folder name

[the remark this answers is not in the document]
Hyodar: No need to rename the branch, this won't change this repository URL (it is from when we merge it to main).

Umee-Bots/Lending-Pool-Reentrancy/src/agent.config.ts
```diff
@@ -0,0 +1,24 @@
...
+    "rebalanceStableBorrowRate(address,address)",
+    "setUserUseReserveAsCollateral(address, bool)",
+    "liquidationCall(address,address,address,uint256,bool)",
+    "finalizeTransfer(address,address,address,uint256,uint256,uint256)",
+  ],
+  // Address of the LendingPool contract
+  lendingPoolAddress: "0xce744a9baf573167b2cf138114ba32ed7de274fa",
+};
+
+// Uncomment these lines for the testnet test:
```
0xtaf, line 21: just grammatical concern: 
// Uncomment `this line` for the testnet test:
if there's only 1 line. otherwise it might feel like something is missing.

0xtaf (approved):
after the last updates it is really good job. 👍

Umee-Bots/Lending-Pool-Reentrancy/src/utils.ts
```diff
@@ -0,0 +1,32 @@
+import { Finding, FindingSeverity, FindingType } from "forta-agent";
+import { encodeFunctionSignature } from "forta-agent-tools/lib/utils";
```
Hyodar, line 2: `forta-agent-tools` library is in `devDependencies` and is used outside of development-related situations.

Umee-Bots/Lending-Pool-Reentrancy/src/agent.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      );
+
+      const expected: Finding[] = [];
+      expected.push(utils.createFinding(testEncodedDepositFuncCall.slice(0, 10), sighashes[1]));
+
+      const findings: Finding[] = await handleTx(tx);
+      expect(findings).toStrictEqual(expected);
+    });
+
+    it("Should detect three finding if deposit,withdraw,deposit,any call, deposit.... any calls then withdraw in a reentrant way", async () => {
```
Hyodar, line 97: This could use some rephrasing.

Hyodar: Also, this test should only result in one finding.

Umee-Bots/Lending-Pool-Reentrancy/src/agent.ts
```diff
@@ -0,0 +1,42 @@
...
+          }
+
+          if (traces[j].action.to === config.lendingPoolAddress && !alerted) {
+            const selector = (traces[j].action.input || "").slice(0, 10); // "0x" and first 4 bytes
+            if (sigHashes.includes(selector)) {
+              const initialCallSelector = (txEvent.transaction.data || "").slice(0, 10);
+              findings.push(utils.createFinding(initialCallSelector, selector));
+              alerted = true;
+            }
+          }
```
Hyodar, line 31:
```suggestion
        for (i = i + 1; i < traces.length; i++) {
          if (traces[i].traceAddress.length <= depth) {
            i--;
            break; // subtree ended, non reentrant call
          }

          if (!alerted && traces[i].action.to === config.lendingPoolAddress) {
            const selector = (traces[i].action.input || "").slice(0, 10); // "0x" and first 4 bytes
            if (sigHashes.includes(selector)) {
              const initialCallSelector = (trace.action.input || "").slice(0, 10);
              findings.push(utils.createFinding(initialCallSelector, selector));
              alerted = true;
            }
          }
```
This fixes the unexpected test result mentioned in a previous comment. Also, about `initialCallSelector`, since the call it's referring to is the one from `trace`, the selector should be from `trace` input, not the transaction data.

Umee-Bots/Lending-Pool-Reentrancy/src/agent.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      const iFace = new Interface([
+        "function withdraw(address asset,uint256 amount,address to)",
+        "function deposit(address asset,uint256 amount,address onBehalfOf,uint16 referralCode)",
+      ]);
+
+      const testEncodedWithdrawFuncCall: string = iFace.encodeFunctionData("withdraw", [
+        createAddress("0x0a"),
+        234,
+        createAddress("0x0b"),
+      ]);
```
Hyodar, line 107: `testEncodedWithdrawFuncCall` is not used anywhere.

Umee-Bots/Lending-Pool-Reentrancy/src/agent.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+          createTrace([0, 0, 0, 0], depositSig), // Reentrant deposit call
+        ],
+        testEncodedDepositFuncCall
+      );
+
+      const expected: Finding[] = [];
+      expected.push(utils.createFinding(testEncodedDepositFuncCall.slice(0, 10), sighashes[1]));
+      expected.push(utils.createFinding(sighashes[3], testEncodedDepositFuncCall.slice(0, 10)));
+      expected.push(utils.createFinding(sighashes[3], testEncodedDepositFuncCall.slice(0, 10)));
+      expected.push(utils.createFinding(testEncodedDepositFuncCall.slice(0, 10), sighashes[3]));
```
Hyodar, line 134: This is not the expected result, only the first deposit call should trigger a finding.

Umee-Bots/Lending-Pool-Reentrancy/src/agent.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+
+      const findings: Finding[] = await handleTx(tx);
+      expect(findings).toStrictEqual(expected);
+    });
+  });
```
Hyodar, line 139: It'd be good to have a test with a more complex call tree. For example, traces that aren't to `LendingPool` mixed with traces to `LendingPool`, traces with more than one subtrace...

Umee-Bots/lendingPool-reentrancy/src/agent.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+        ],
+        testEncodedWithdrawFuncCall
+      );
+
+      const expected: Finding[] = [];
+      expected.push(utils.createFinding(testEncodedWithdrawFuncCall, sighashes[0]));
+      expected.push(utils.createFinding(testEncodedWithdrawFuncCall, sighashes[0]));
+
+      const findings: Finding[] = await handleTx(tx);
+      expect(findings.length).toStrictEqual(expected.length);
```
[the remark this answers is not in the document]
Hyodar: Unresolving this conversation.

Hyodar (changes requested):

Hyodar (approved):
Did some considerable changes.

Umee-Bots/Lending-Pool-Reentrancy/src/agent.ts
```diff
@@ -1,31 +1,54 @@
...
+            if (func && config.reentrancyBlacklist.includes(func.name)) {
+              let initialFunc;
+
+              try {
+                initialFunc = lendingPool.getFunction(trace.action.input.slice(0, 10));
+              } catch {
+                initialFunc = { name: "(unknown)" };
+              }
+
+              if (initialFunc) {
```
e1Ru1o, line 48: this if seems to not be needed

Umee-Bots/Lending-Pool-Reentrancy/src/agent.spec.ts
```diff
@@ -9,24 +9,34 @@ import {
...
-      to: CONFIG.lendingPoolAddress,
+      to,
       input,
     } as TraceAction,
   } as Trace;
 };
 
+const CONFIG: AgentConfig = {
+  reentrancyBlacklist: ["withdraw", "deposit"],
+  lendingPoolAddress: createAddress("4001"),
```
e1Ru1o, line 33: missing 0x here

Umee-Bots/Lending-Pool-Reentrancy/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "push": "forta-agent push",
+    "disable": "forta-agent disable",
+    "enable": "forta-agent enable",
+    "keyfile": "forta-agent keyfile",
+    "test": "jest --detectOpenHandles",
+    "format": "prettier --write \"src/**/*.ts\""
+  },
+  "dependencies": {
+    "forta-agent": "^0.1.6",
+    "forta-agent-tools": "^2.0.6"
```
e1Ru1o, line 28: should be a devDep

Umee-Bots/Lending-Pool-Reentrancy/src/agent.ts
```diff
@@ -0,0 +1,61 @@
...
+      if (trace.action.to === lendingPoolAddress) {
+        let alerted = false;
+
+        for (i = i + 1; i < traces.length; i++) {
+          if (traces[i].traceAddress.length <= depth) {
+            i--;
+            break; // subtree ended
+          }
+
+          if (!alerted && traces[i].action.to === config.lendingPoolAddress) {
```
e1Ru1o, line 29:
```suggestion
          if (!alerted && traces[i].action.to === lendingPoolAddress) {
```
Change needed

Umee-Bots/Lending-Pool-Reentrancy/src/agent.spec.ts
Hyodar: Added a different test for those

PR #250, opened 2022-05-30
TJ03 - Large Swap

RCantu92:
# Large Swap Bot

## Types of changes

- [x] New Bot
- [ ] Fix a bot error

## New Bot PR readiness

- [x] Bot contains tests

## Comments (optional)


TraderJoe-Bots/Large-Swap/src/agent.ts
```diff
@@ -0,0 +1,98 @@
...
+        // Assign the `Swap` arguments to variables
+        const [, amount0In, amount1In, amount0Out, amount1Out] = log.args;
+
+        // Create multicall2 instance based on
+        // the current provider's chain id
+        const multicall2: MulticallProvider = new MulticallProvider(
+          provider,
+          networkManager.multicall2Data,
+          networkManager.chainId
+        );
```
Hyodar, line 38: Suggestion: this could be created only once in `provideHandleTransaction` to avoid creating it on every log or every transaction.

TraderJoe-Bots/Large-Swap/src/agent.ts
```diff
@@ -0,0 +1,98 @@
...
+
+        // Create pair contract at emitting address
+        // and fetch `token0` and `token1`
+        const pairContract = new MulticallContract(log.address, PAIR_ABI);
+        const tokenResults = await multicall2.all([pairContract.token0(), pairContract.token1()], txEvent.blockNumber);
```
Hyodar, line 43: Maybe this multicall here could be on `txEvent.blockNumber - 1`, then it could also include the `getReserves` call.

TraderJoe-Bots/Large-Swap/src/multicall2.ts
```diff
@@ -0,0 +1,65 @@
...
+
+class MulticallProvider extends Provider {
+  constructor(provider: EthersProvider, multicallAddress: Record<number, string>, chainId?: number) {
+    // clean ethers-multicall Multicall addresses list so a Multicall contract is not called by mistake
+    const ethersMulticallChains = [1, 3, 4, 5, 42, 56, 66, 97, 100, 128, 137, 250, 1337, 42161, 43114, 80001];
+    ethersMulticallChains.forEach((network) => {
+      setMulticallAddress(network, undefined as unknown as string);
+    });
+    for (const [network, address] of Object.entries(multicallAddress)) {
+      // setMulticallAddress(network as unknown as number, address);
```
Hyodar, line 18:
```suggestion
```

TraderJoe-Bots/Large-Swap/src/agent.spec.ts
```diff
@@ -0,0 +1,452 @@
...
+    requireSuccess: boolean,
+    calls: [string, string][],
+    returnData: [boolean, string][],
+    blockNumber: number
+  ) => {
+    mockProvider.addCallTo(multicall2Address, blockNumber, MULTICALL2_IFACE, "tryAggregate", {
+      inputs: [requireSuccess, calls],
+      outputs: [returnData],
+    });
+  };
```
Hyodar, line 142: I think it's fine to keep it the way it is in this case, but there's a way to test multicalls that's currently being used in UMEE #7 that might be easier (this is for `aggregate`, but adapting it to `tryAggregate` is straightforward):

https://github.com/NethermindEth/Forta-Agents/blob/845793101ba001a12acf075ee88f6b4a3ab79f55/Umee-Bots/High-Utilization-Ratio/src/agent.spec.ts#L46-L66

https://github.com/NethermindEth/Forta-Agents/blob/845793101ba001a12acf075ee88f6b4a3ab79f55/Umee-Bots/High-Utilization-Ratio/src/agent.spec.ts#L154-L159

This basically replaces the mock `call` from `mockProvider`, adding a sort of a proxy that decodes and makes the multicall calls one by one using `mockMulticallProvider` and encodes the results if the original call was made to the multicall address, or just redirects the call to `mockProvider`'s `call` otherwise.

Then, instead of having to specify every multicall input and output, you can just specify the multicall call results individually using, for example, `mockMulticallProvider.addCallTo`.

TraderJoe-Bots/Large-Swap/src/multicall2.ts
```diff
@@ -0,0 +1,65 @@
...
+        break;
+      }
+
+      const params = Abi.decode(outputs, result.returnData);
+      results.push({
+        success: result.success,
+        returnData: outputs.length === 1 ? params[0] : params,
+      });
+    }
+    return results as T;
```
e1Ru1o, line 61: I think in this case I prefer to use `true` when calling `tryAggregate` at the moment I have no case in mind when I would like to use `false` because from the bots perspective, each time we query the blockchain is because we need some info. This will imply to have a `try/catch` in this section of code.

Another thing is that I would like to return follow the `trySomething` pattern, which return `[bool, ...returnedTypes]` so the first element in the returned data indicates if everything was ok or not. So when doing test calls like fetching tokens as we do in this bot we can have something like:
```typescript
const [success, ...results] = await multicall2.all(listOfCalls, blockNumber);

if(success) {
  // results contains all the data needed
  doChecks(results);
}
else {
  // something fails in the calls
}
```

e1Ru1o: cc: @Hyodar, wdyt?

Hyodar: Yes, about using `true` on `tryMulticall`, I was currently working on adding the normal `aggregate` to this `Multicall2`, then separating `all` and `tryAll`. In the former, only the actual call results are returned, so I think it's better in terms of usability when there's no need for the `success` flag.

About the output pattern, we could work on something like that, but `tryAggregate` returns a flag of success for each of the calls, and that could be useful in some cases. So, another option we have that is a bit different from the current approach is returning an array of `success` and an array of `results`.

Hyodar: Also, I don't think this function should stop decoding the results if one call is not successful. Whoever is using it should be dealing with `success` later, since one of the main use cases is grouping calls that can fail.

e1Ru1o: Agree with the above comment. Just another note on my suggestion, I know they are returning that `success` flag per call, but the flag I would like to have is to be able to know if something fails, is like having a short circuit flag.

But yes, when success is required it is better to use `aggregate` so having that `all` and `tryAll` would be nice.

I think we need to be agree on what to change in this PR and what left as it is, because I don't want to jeopardize the delivery with technicalities.

e1Ru1o: Or we can keep it as it is this time, because it is working :upside_down_face:, and what we are discussing is how we would like to build this to use it from now on. We can do that later, and maybe add it to GAM.

RCantu92: ☝️ This is what I was thinking as well. In general, I want to study the `multicall`/`multicall2` approach more, so I have a firmer understanding of how it works, since it is new to me. (I feel I have a good basic understanding, but I'd like to understand it more)

Then that could inform me on how to make it more "generalized" for other use cases. I figured my current implementation works for the bot I currently have, but it would need to be updated and worked on before adding to the `GAM`.

TraderJoe-Bots/Large-Swap/src/agent.ts
```diff
@@ -0,0 +1,112 @@
...
+
+        // Check if the emitting address is a valid pair contract
+        // by comparing to `create2` output.
+        const create2PairAddress: string = create2Pair(
+          token0.toLowerCase(),
+          token1.toLowerCase(),
+          networkManager.factory,
+          networkManager.pairInitCodeHash
+        );
+        if (create2PairAddress === log.address && reserveSuccess === true) {
```
Hyodar, line 86:
```suggestion
  // Create multicall2 instance based on
  // the current provider's chain id
  const multicall2: MulticallProvider = new MulticallProvider(
    provider,
    networkManager.multicall2Data,
    networkManager.chainId
  );

  return async (txEvent: TransactionEvent): Promise<Finding[]> => {
    const findings: Finding[] = [];

    await Promise.all(
      // Filter for any `Swap` event emissions
      txEvent.filterLog(SWAP_ABI).map(async (log) => {
        // Assign the `Swap` arguments to variables
        const [, amount0In, amount1In, amount0Out, amount1Out] = log.args;

        // Create pair contract at emitting address
        // and fetch `token0` and `token1`
        const pairContract = new MulticallContract(log.address, PAIR_ABI);

        const results = await multicall2.all(
          [pairContract.token0(), pairContract.token1(), pairContract.getReserves()],
          txEvent.blockNumber - 1
        );

        if (results.some((el) => !el.success)) {
          return;
        }

        const [token0, token1, [reserve0, reserve1]] = results.map((el) => el.returnData);

        // Check if the emitting address is a valid pair contract
        // by comparing to `create2` output.
        const create2PairAddress: string = create2Pair(
          token0.toLowerCase(),
          token1.toLowerCase(),
          networkManager.factory,
          networkManager.pairInitCodeHash
        );
        if (create2PairAddress === log.address) {
```

TraderJoe-Bots/Large-Swap/src/multicall2.ts
```diff
@@ -0,0 +1,64 @@
...
+
+      if (result.success === false) {
+        continue;
+      }
+
+      const params = Abi.decode(outputs, result.returnData);
+      results.push({
+        success: result.success,
+        returnData: outputs.length === 1 ? params[0] : params,
+      });
```
Hyodar, line 58:
```suggestion
      if (result.success) {
        const params = Abi.decode(outputs, result.returnData);
        results.push({
          success: result.success,
          returnData: outputs.length === 1 ? params[0] : params,
        });
      }
      else {
        results.push({
          success: false,
          returnData: null,
        });
      }
```
Considering what we discussed about still returning the call result when it's not sucessful, this suggestion does this.

Simply removing the `result.success` condition that leads to a `continue` wouldn't be enough, because if the call is not successful, trying and decoding the return data will lead to an error. So, considering that, it's necessary to check if it was successful or not before decoding.

TraderJoe-Bots/Large-Swap/src/agent.ts
```diff
@@ -0,0 +1,112 @@
...
+    );
+
+    return findings;
+  };
+}
+
+export default {
+  initialize: provideInitialize(getEthersProvider()),
+  handleTransaction: provideHandleTransaction(getEthersProvider(), networkManager, THRESHOLD_PERCENTAGE),
+};
```
Hyodar, line 112:
```suggestion
import { Finding, HandleTransaction, TransactionEvent, getEthersProvider } from "forta-agent";
import { providers, BigNumber } from "ethers";
import NetworkManager from "./network";
import NetworkData from "./network";
import { SWAP_ABI, PAIR_ABI, create2Pair } from "./utils";
import { createFinding } from "./findings";
import { MulticallContract, MulticallProvider } from "./multicall2";

const THRESHOLD_PERCENTAGE: number = 20;

let networkManager: NetworkData;
let multicall2: MulticallProvider;

export const provideInitialize = (_networkManager: NetworkData, provider: providers.Provider) => async () => {
  const { chainId } = await provider.getNetwork();
  
  networkManager = _networkManager;
  networkManager.setNetwork(chainId);

  multicall2 = new MulticallProvider(provider, networkManager.multicall2Data, chainId);
};

export function provideHandleTransaction(thresholdPercentage: number): HandleTransaction {
  return async (txEvent: TransactionEvent): Promise<Finding[]> => {
    const findings: Finding[] = [];

    await Promise.all(
      // Filter for any `Swap` event emissions
      txEvent.filterLog(SWAP_ABI).map(async (log) => {
        // Assign the `Swap` arguments to variables
        const [, amount0In, amount1In, amount0Out, amount1Out] = log.args;

        // Create pair contract at emitting address
        // and fetch `token0` and `token1`
        const pairContract = new MulticallContract(log.address, PAIR_ABI);

        const results = await multicall2.all(
          [pairContract.token0(), pairContract.token1(), pairContract.getReserves()],
          txEvent.blockNumber - 1
        );

        if (results.some((el) => !el.success)) {
          return;
        }

        const [token0, token1, [reserve0, reserve1]] = results.map((el) => el.returnData);

        // Check if the emitting address is a valid pair contract
        // by comparing to `create2` output.
        const create2PairAddress: string = create2Pair(
          token0.toLowerCase(),
          token1.toLowerCase(),
          networkManager.factory,
          networkManager.pairInitCodeHash
        );
        if (create2PairAddress === log.address) {
          // Create threshold amounts
          const reserve0Threshold: BigNumber = reserve0.mul(thresholdPercentage).div(100);
          const reserve1Threshold: BigNumber = reserve1.mul(thresholdPercentage).div(100);

          // If the `amount` arguments from `Swap` are large relative
          // to their respective `reserve`, create a finding
          if (
            amount0In.gte(reserve0Threshold) ||
            amount0Out.gte(reserve0Threshold) ||
            amount1In.gte(reserve1Threshold) ||
            amount1Out.gte(reserve1Threshold)
          ) {
            findings.push(createFinding(log.args, log.address));
          }
        }
      })
    );

    return findings;
  };
}

export default {
  initialize: provideInitialize(new NetworkManager(), getEthersProvider()),
  handleTransaction: provideHandleTransaction(THRESHOLD_PERCENTAGE),
};
```
Just a refactoring suggestion - some changes to allow creating a `MulticallProvider` only once while allowing unit testing with a mock network manager and some changes to the way the `tryAggregate` return is handled.

TraderJoe-Bots/Large-Swap/src/agent.spec.ts
```diff
@@ -0,0 +1,512 @@
+import { FindingType, FindingSeverity, Finding, HandleTransaction, TransactionEvent } from "forta-agent";
+import { TestTransactionEvent, createAddress, MockEthersProvider, MockEthersSigner } from "forta-agent-tools/lib/tests";
```
Hyodar, line 2:
```suggestion
import { TestTransactionEvent, createAddress, MockEthersProvider } from "forta-agent-tools/lib/tests";
```
Unused imported module.

TraderJoe-Bots/Large-Swap/src/agent.spec.ts
```diff
@@ -0,0 +1,512 @@
...
+import { BigNumber } from "ethers";
+import { Interface, keccak256, getCreate2Address } from "ethers/lib/utils";
+import { provideHandleTransaction } from "./agent";
```
Hyodar, line 5:
```suggestion
import { provideHandleTransaction, provideInitialize } from "./agent";
```
Related to `agent.ts` refactor.

TraderJoe-Bots/Large-Swap/src/agent.spec.ts
```diff
@@ -0,0 +1,512 @@
...
+  ) => {
+    mockProvider.addCallTo(multicall2Address, blockNumber, MULTICALL2_IFACE, "tryAggregate", {
+      inputs: [requireSuccess, calls],
+      outputs: [returnData],
+    });
+  };
+
+  beforeEach(() => {
+    mockProvider.clear();
+  });
```
Hyodar, line 146:
```suggestion
  let handleTransaction: HandleTransaction;

  const createTryAggregateCall = (
    multicall2Address: string,
    requireSuccess: boolean,
    calls: [string, string][],
    returnData: [boolean, string][],
    blockNumber: number
  ) => {
    mockProvider.addCallTo(multicall2Address, blockNumber, MULTICALL2_IFACE, "tryAggregate", {
      inputs: [requireSuccess, calls],
      outputs: [returnData],
    });
  };

  beforeAll(() => {
    handleTransaction = provideHandleTransaction(testThresholdPercentage);
  });

  beforeEach(async () => {
    mockProvider.clear();

    // @ts-expect-error
    mockProvider.getNetwork = jest.fn().mockReturnValue({ chainId: mockNetworkManager.chainId });

    const initialize = provideInitialize(mockNetworkManager, mockProvider as any);
    await initialize();
  });
```
Related to the `agent.ts` refactor.

Hyodar:
Great work! These are the comments I had for now.

TraderJoe-Bots/Large-Swap/src/agent.spec.ts
```diff
@@ -0,0 +1,516 @@
...
+  beforeAll(() => {
+    handleTransaction = provideHandleTransaction(testThresholdPercentage);
+  });
+
+  beforeEach(async () => {
+    mockProvider.clear();
+    // @ts-expect-error
+    mockProvider.getNetwork = jest.fn().mockReturnValue({ chainId: mockNetworkManager.chainId });
+    const initialize = provideInitialize(mockNetworkManager, mockProvider as any);
+    await initialize();
```
yasmine9, line 149: It's better to avoid calling `initialize` inside tests. As mentioned in the previous comment, this can be avoided If the `networkManager` and `multicall2` are passed as parameters.

TraderJoe-Bots/Large-Swap/src/network.ts
```diff
@@ -0,0 +1,61 @@
+interface NetworkData {
+  chainId: number;
+  multicall2: string;
```
yasmine9, line 3: Why `multicall2` is needed? the address is stored inside `multicall2Data` too which seems to be the one used in the bot logic.

yasmine9 (changes requested):
Great Job Roberto 👏 , I can approve after some comments are re-checked.

Hyodar (approved):
Bot logic seems great :+1:

PR #252, opened 2022-05-31
UM06 - Large Borrow

Hyodar:
# Large Borrow Bot

## Types of changes

- [x] New Bot
- [ ] Fix a bot error

## New Bot PR readiness

- [x] Bot contains tests


yehia67 (approved):

0xtaf (approved):
LGTM 🚀

Umee-Bots/Large-Borrow/src/agent.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  return async (txEvent: TransactionEvent): Promise<Finding[]> => {
+    const findings: Finding[] = [];
+    const logs = txEvent.filterLog(BORROW_ABI, config.lendingPoolAddress);
+
+    await Promise.all(
+      logs.map(async (log) => {
+        const { reserve, amount, user, onBehalfOf } = log.args;
+
+        const { uTokenAddress } = await LendingPool.getReserveData(reserve, { blockTag: "latest" });
```
e1Ru1o, line 29: why `"latest"` here?

Umee-Bots/Large-Borrow/src/utils.ts
```diff
@@ -0,0 +1,112 @@
...
+      };
+    });
+  }
+
+  protected static async smartCall(
+    contract: ethers.Contract,
+    functionName: string,
+    args: any[],
+    overrides: CallOverrides = {},
+    options: SmartCallerOptions
```
e1Ru1o, line 80: this should not work :thinking:, before a default param all the other params needs to have a default value too

Hyodar: Yes, I added `options` after `overrides` and didn't notice that - but it does work 🤔 . Changed now, thanks!

Umee-Bots/Large-Borrow/src/agent.ts
Hyodar: Well, it shouldn't matter a lot here to use the block number or `"latest"`, because this shouldn't change (internally, `SmartCall` should be frequently getting it from the cache).

e1Ru1o: Ok, because lendingPool is not caching by blockTag it is being cached. :+1:

Umee-Bots/Large-Borrow/src/agent.ts
```diff
@@ -0,0 +1,48 @@
...
+        const erc20Asset = SmartCaller.from(new ethers.Contract(reserve, [BALANCE_OF_ABI], provider));
+        const tvl = await erc20Asset.balanceOf(uTokenAddress, { blockTag: txEvent.blockNumber - 1 });
+
+        const percentage = new BigNumber(amount.toString()).div(new BigNumber(tvl.toString())).shiftedBy(2);
```
yehia67, line 33: `amount` and `tvl` are already `BigNumber` why not using `.div` and `.shiftedBy(2)` directly? 
```
amount.div(tvl).shl(2)
```

Hyodar: No, because `ethers.BigNumber` doesn't support decimals directly. So, to do that with `ethers.BigNumber`, we'd need to choose a precision and adapt the division to use it so the final result would be, e.g. `99` instead of `0.99` which would be truncated to `0` (the `shl` would also need to be before the division).

`BigNumber` makes this easier and more straightforward because it supports decimals directly, so just a simple division works and it will automatically show the decimal point when `bn.toString()` is called.

PR #259, opened 2022-06-03
UM07 - High Utilization Ratio

Hyodar:
# High Utilization Ratio Bot

## Types of changes

- [x] New Bot
- [ ] Fix a bot error

## New Bot PR readiness

- [x] Bot contains tests


Umee-Bots/High-Utilization-Ratio/src/agent.ts
```diff
@@ -0,0 +1,150 @@
...
+
+const reserveData: ReserveData[] = [];
+
+// testing
+export const resetReserveData = () => (reserveData.length = 0);
+export const getReserveData = () => reserveData;
+export const setReserveData = (newReserveData: ReserveData[]) => {
+  resetReserveData();
+  newReserveData.forEach((el) => reserveData.push(el));
+};
```
0xtaf, line 36: I think this block should be inside agent.spec.ts since they're only related to testing

Umee-Bots/High-Utilization-Ratio/src/agent.ts
```diff
@@ -0,0 +1,150 @@
...
+
+let multicallProvider: MulticallProvider;
+
+export const provideInitialize = (provider: ethers.providers.Provider, config: AgentConfig): Initialize => {
+  return async () => {
+    const LendingPool = new ethers.Contract(
```
0xtaf, line 42: why not camelCase?

0xtaf, line 42: why not camelCase?

Umee-Bots/High-Utilization-Ratio/src/agent.ts
```diff
@@ -0,0 +1,150 @@
...
+
+        if (notOnCooldown) {
+          findings.push(createAbsoluteThresholdFinding(reserve.asset.address, currentUsageRatio));
+          reserve.lastAlertTimestamp.absolute = timestamp;
+        }
+      }
+
+      const percentageIncrease =
+        currentUsageRatio.isZero() && lastUsageRatio.isZero()
+          ? new BigNumber(0)
```
0xtaf, line 125:
```suggestion
          ? new BigNumber("0")
```
is it better to make this string?

0xtaf, line 125:
```suggestion
          ? new BigNumber("0")
```
is it better to make this string?

0xtaf (approved):
Great work again!

Umee-Bots/High-Utilization-Ratio/src/agent.ts
Hyodar: Yes, but the idea in this case is exporting functions to access the bot local state. This could be changed to just exporting `reserveData`, but I don't know if it's worth it to change this now.

Umee-Bots/High-Utilization-Ratio/src/agent.ts
Hyodar: This is technically a style rule break, but I do think it looks better in this case to use PascalCase since that's the way it's used in solidity and ethers basically mimicks that.

Umee-Bots/High-Utilization-Ratio/src/agent.ts
Hyodar: Good catch, but I do think it's fine in this case, because `0` shouldn't incur in a precision error somehow even if it's stored as a float, and it should be a tiny bit faster to parse this way.

Umee-Bots/High-Utilization-Ratio/src/agent.ts
e1Ru1o: I prefer to follow the same style everywhere

Umee-Bots/High-Utilization-Ratio/src/agent.spec.ts
```diff
@@ -0,0 +1,560 @@
...
+        }>;
+        const results = calls.map((call) =>
+          mockMulticallProvider.call({ data: call.callData, to: call.target }, blockTag)
+        );
+        return MULTICALL_IFACE.encodeFunctionResult("aggregate", [ethers.BigNumber.from(blockTag), results]);
+      } else {
+        return _call({ data, to, from });
+      }
+    });
+  }
```
e1Ru1o, line 66: nice :ok_hand:

Umee-Bots/High-Utilization-Ratio/src/agent.ts
```diff
@@ -0,0 +1,150 @@
...
+          reserveAddresses[idx],
+          data.uTokenAddress,
+          data.stableDebtTokenAddress,
+          data.variableDebtTokenAddress
+        )
+      );
+    });
+
+    multicallProvider = new MulticallProvider(provider);
+    await multicallProvider.init();
```
e1Ru1o, line 63: I would move the assignment (L62) to L38 where the variable is defined and keep the `init` call here. I would add a `multicallProvider` parameter to all the `provideHandle...` so in tests you can pass the provider. 

This change along with the one for `reserveData` will allow you to remove many initialize calls you have in tests.

Umee-Bots/High-Utilization-Ratio/src/agent.ts
```diff
@@ -0,0 +1,150 @@
...
+          reserveInit.args.uToken,
+          reserveInit.args.stableDebtToken,
+          reserveInit.args.variableDebtToken
+        )
+      );
+    });
+
+    return [];
+  };
+};
```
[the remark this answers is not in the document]
e1Ru1o: You can keep it this way in this bot, but keep getLogs in mind for future bots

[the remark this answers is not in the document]
Hyodar: Right. I agree, but now that I think about it, isn't this something we should ask for Forta SDK to be making available? If they already fetched all events from every transaction, it seems like they could just get all of those and add it to `BlockEvent`.

Umee-Bots/High-Utilization-Ratio/src/agent.ts
```diff
@@ -0,0 +1,150 @@
...
+    const multicalls = arrayChunks(reserveData, 10).map((chunk) => {
+      return multicallProvider.all(
+        chunk.flatMap((reserve) => [
+          reserve.asset.balanceOf(reserve.uTokenAddress),
+          reserve.stableDebtToken.totalSupply(),
+          reserve.variableDebtToken.totalSupply(),
+        ]),
+        blockEvent.blockNumber
+      );
+    });
```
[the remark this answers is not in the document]
Hyodar: Yes! That's because providers do have "gas limits" for calls, so it's better to try and make sure the call won't fail because of that.

By the way, I did a stress test on both this bot and UM01 with the Alchemy providers, and they both seemed to work fine at max chunk capacity.

Umee-Bots/High-Utilization-Ratio/src/agent.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    // divide into chunks of 3 later since each reserve needs 3 calls
+    const data = arrayChunks((await Promise.all(multicalls)).flat(), 3) as ReserveBalances[];
+
+    const usageRatios = data.map((el) => usageRatio(...el));
```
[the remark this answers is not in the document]
Hyodar: 👍

Umee-Bots/High-Utilization-Ratio/src/agent.ts
Hyodar: Done

Umee-Bots/High-Utilization-Ratio/src/agent.ts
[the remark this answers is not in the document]
Hyodar: Really? Don't they get all the info in a node at once and then distribute it to the all the bots that are running there? (I don't know, asking for real)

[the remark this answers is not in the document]
e1Ru1o: Oh not sure about that, maybe we can take a look at the sdk to get some insight

Umee-Bots/High-Utilization-Ratio/src/agent.spec.ts
```diff
@@ -0,0 +1,557 @@
...
+    });
+
+    it("should emit a finding if a reserve's usage ratio is greater than or equal to an absolute threshold", async () => {
+      const config: AgentConfig = {
+        ...DEFAULT_CONFIG,
+        absoluteThreshold: "0.7",
+        alertCooldown: { absolute: 0, percentage: 0 },
+      };
+
+      initialize = provideInitialize(reserveData, provider, config);
```
e1Ru1o, line 288: I think this line is not needed and the same applies to all the other tests

PR #269, opened 2022-06-13
GMX-05

ericmartihaynes:
# GMX-05

## Types of changes

- [x] New Bot
- [ ] Fix a bot error

## New Bot PR readiness

- [x] Bot contains tests

## Comments (optional)


GMX-Bots/gmx5-detect-frontruns/src/agent.ts
```diff
@@ -0,0 +1,204 @@
...
+          else {
+            callHistory.set(accountBack + token0 + token1, [swapEvent, callHistoryIndex, txEvent.hash]);
+            callHistoryKeys[callHistoryIndex] = accountBack + token0 + token1;
+          }
+        }
+        //transaction gets stored
+        else {
+          callHistory.set(accountBack + token0 + token1, [swapEvent, callHistoryIndex, txEvent.hash]);
+          callHistoryKeys[callHistoryIndex] = accountBack + token0 + token1;
+        }
```
e1Ru1o, line 137:
```suggestion
        //transaction gets stored
        if (!sandwichAttack) {
          callHistory.set(accountBack + token0 + token1, [swapEvent, callHistoryIndex, txEvent.hash]);
          callHistoryKeys[callHistoryIndex] = accountBack + token0 + token1;
        }
```

GMX-Bots/gmx5-detect-frontruns/src/agent.ts
```diff
@@ -0,0 +1,204 @@
...
+      });
+    }
+    return findings;
+  };
+
+// const handleBlock: HandleBlock = async (blockEvent: BlockEvent) => {
+//   const findings: Finding[] = [];
+//   // detect some block condition
+//   return findings;
+// }
```
e1Ru1o, line 199: remove this

GMX-Bots/gmx5-detect-frontruns/src/agent.ts
```diff
@@ -0,0 +1,204 @@
...
+
+export default {
+  handleTransaction: provideHandleTx(GMX_ROUTER_ADDRESS, SWAP_EVENT, getEthersProvider()),
+  // handleBlock
```
e1Ru1o, line 203: remove this

GMX-Bots/gmx5-detect-frontruns/src/agent.spec.ts
```diff
@@ -0,0 +1,438 @@
...
+import { Interface } from "@ethersproject/abi";
+
+const ABI: string[] = [
+  "event Swap(address account, address tokenIn, address tokenOut, uint256 amountIn, uint256 amountOut)",
+];
+
+const TEST_IFACE: Interface = new Interface([
+  ...ABI,
+  "event Swap(address account, address tokenIn, address tokenOut, uint256 amountIn, uint256 amountOut)",
+]);
```
e1Ru1o, line 23:
```suggestion
const TEST_IFACE: Interface = new Interface(ABI);
```

GMX-Bots/gmx5-detect-frontruns/src/agent.spec.ts
```diff
@@ -0,0 +1,438 @@
...
+  "event Swap(address account, address tokenIn, address tokenOut, uint256 amountIn, uint256 amountOut)";
+const handler = provideHandleTx(
+  MOCK_GMX_ROUTER_ADDRESS,
+  MOCK_SWAP_EVENT,
+  mockProvider as unknown as ethers.providers.Provider
+);
+
+describe("sandwich attack frontrun agent", () => {
+  let handleTransaction: HandleTransaction;
+  const mockTxEvent = createTransactionEvent({} as any);
```
e1Ru1o, line 37: remove this because it is not used

e1Ru1o (changes requested):

GMX-Bots/gmx5-detect-frontruns/.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Auto detect text files and perform LF normalization
```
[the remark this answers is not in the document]
ericmartihaynes: No idea how it got there, removed

GMX-Bots/gmx5-detect-frontruns/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Changing Network
+By default this bot uses the Arbitrum network, but it can be changed to Avalanche by setting "jsonRpcUrl" and "traceRpcUrl" to
+https://api.avax.network/ext/bc/C/rpc in the forta.config.json file. The GMX_ROUTER_ADDRESS in agent.ts must be changed to "0x5F719c2F1095F7B9fc68a68e35B51194f4b6abe8".
+
+## Test Data
+
+The bot behaviour can be verified with the following transactions:
+
+- ?????????????????????????????
```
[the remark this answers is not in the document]
ericmartihaynes: I don't have any transactions to test with, should I just remove this section?

GMX-Bots/gmx5-detect-frontruns/src/agent.ts
```diff
@@ -0,0 +1,205 @@
...
+          );
+        }
+      });
+    }
+    return findings;
+  };
+
+export default {
+  initialize: initialize(provider),
+  handleTransaction: provideHandleTx(GMX_ROUTER_ADDRESS, SWAP_EVENT, provider),
```
e1Ru1o, line 204: this is not going to work, please do the following experiment:

1. Add a `console.log` in the `initialize` function to display the address you get from the `NetworkManager`.
2. Add a `console.log` in the line 50 for example to display `router`.
3. Do `npm run` on one of the supported networks and see what is printed.

Reach me on slack after that to discuss

e1Ru1o (changes requested):

GMX-Bots/gmx5-detect-frontruns/src/agent.ts
```diff
@@ -0,0 +1,206 @@
...
+              if (
+                (event[1] > potentialSandwhichFront[1] &&
+                  event[1] < callHistoryIndex &&
+                  callHistoryIndex > potentialSandwhichFront[1]) ||
+                (event[1] < potentialSandwhichFront[1] &&
+                  event[1] < callHistoryIndex &&
+                  callHistoryIndex < potentialSandwhichFront[1]) ||
+                (event[1] > potentialSandwhichFront[1] &&
+                  event[1] > callHistoryIndex &&
+                  callHistoryIndex < potentialSandwhichFront[1])
```
yasmine9, line 121: I don't get what are you exactly checking in here. Can you explain ? 
I tried to understand this from a logic pov and i found out that: 

1.  **The first check:** `(event[1] > potentialSandwhichFront[1] &&  event[1] < callHistoryIndex &&  callHistoryIndex > potentialSandwhichFront[1])` is simply equivalent to 
`event[1] > potentialSandwhichFront[1] &&  event[1] < callHistoryIndex`. ie. the 3rd condition is not needed because 1+2 implies 3. 

2. **The second check:** `(event[1] < potentialSandwhichFront[1] && event[1] < callHistoryIndex && callHistoryIndex < potentialSandwhichFront[1])` is equivalent to 
`event[1] < callHistoryIndex && callHistoryIndex < potentialSandwhichFront[1]`. ie. the first condition is not needed because 2+3=> 1 . 
   
3. **The third check:**  is equivalent to `event[1] > potentialSandwhichFront[1] && callHistoryIndex < potentialSandwhichFront[1]`. ie the 2nd condition is not needed. 

Now overall it seems like you are checking if : 

```
potentialSandwhichFront[1] < event[1] < callHistoryIndex ||
event[1] < callHistoryIndex < potentialSandwhichFront[1] ||
callHistoryIndex < potentialSandwhichFront[1] < event[1]
```
I don't think this is what you intended to do, can you please explain this part.

GMX-Bots/gmx5-detect-frontruns/src/agent.ts
```diff
@@ -0,0 +1,206 @@
...
+                  victimTokenIn = tokenInVictim;
+                  victimTokenOut = tokenOutVictim;
+                  victimTokenInAmount = amountInVictim;
+                  victimTokenOutAmount = amountOutVictim;
+                  frontrunnerAddress = accountFront;
+                  frontrunnerProfit = amountOutBack - amountInFront;
+                  victimAddressAndTokens = addressAndTokens;
+                  frontrunnerAddressAndTokens = accountBack + token0 + token1;
+                }
+              }
```
yasmine9, line 138: Why you decided to store the values then check again if there's a frontRun later. 

It would be more organized to have something like this pattern:
```
if ( victim condition checked) {
// set sandwichAttack to true
// delete both victim and frontrunner from history 
// create finding using the the event args. ( no need to store them somewhere else) 
}
```

GMX-Bots/gmx5-detect-frontruns/src/agent.ts
```diff
@@ -0,0 +1,206 @@
...
+          }
+        }
+
+        if (!sandwichAttack) {
+          callHistory.set(accountBack + token0 + token1, [swapEvent, callHistoryIndex, txEvent.hash]);
+          callHistoryKeys[callHistoryIndex] = accountBack + token0 + token1;
+        }
+
+        callHistoryIndex++;
+        mapCleanerCounter++;
```
yasmine9, line 149: Why the Index is incremented even if a sandwichAttack is detected?

GMX-Bots/gmx5-detect-frontruns/src/agent.ts
```diff
@@ -0,0 +1,206 @@
...
+};
+let GMX_ROUTER_ADDRESS = { value: "" };
+const provider = getEthersProvider();
+export const initialize = (provider: ethers.providers.Provider) => async () => {
+  const networkManager = new NetworkManager(data, (await provider.getNetwork()).chainId);
+  GMX_ROUTER_ADDRESS.value = networkManager.get("address");
+};
+export const SWAP_EVENT =
+  "event Swap(address account, address tokenIn, address tokenOut, uint256 amountIn, uint256 amountOut)";
+
```
yasmine9, line 37: I'd recommend moving the network data, ABIs and some constants ( `CALL_HISTORY_SIZE` for example) to a separate file.

yasmine9 (changes requested):

GMX-Bots/gmx5-detect-frontruns/src/agent.ts
```diff
@@ -0,0 +1,206 @@
...
+  HandleTransaction,
+  TransactionEvent,
+  FindingSeverity,
+  FindingType,
+  Log,
+  LogDescription,
+  ethers,
+  getEthersProvider,
+} from "forta-agent";
+import { NetworkManager } from "forta-agent-tools";
```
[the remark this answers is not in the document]
ericmartihaynes: Removed theProvider. As for the `networkManager`, I had it the way you suggested before, but had to change it because it was passed to `provideHandleTx` by value and not by reference, so the changes that happened in `initialize` didn't do anything. Lázaro suggested wrapping it in an object as a solution to this problem

GMX-Bots/gmx5-detect-frontruns/src/agent.ts
ericmartihaynes: Some of the conditions are redundant like you pointed out, but I thought it made more sense to leave them like this for the sake of understanding. Basically this condition checks if the victim transaction is in the sandwich, but because we have an array of a limited size that is similar to a circular array, we can't just check if the index of the victim transaction is in between the front and the back one (because some transaction could find the array full and occupy an earlier index). 
So there are 3 possible situations, where in each `potentialSandwhichFront[1]` is the front transaction, `event[1]` is the victim transaction, and `callHistoryIndex` is the back transaction:

1. In the first case (assuming array size 20), we check that the victim transaction is after the front one, the victim one is before the back one, and the back one is after the front one. Some example values can be `4, 6, 9`
2. In the second case (assuming array size 20), we check that the victim transaction is before the front transaction, the victim one is before the back one, and the back one is before the front. Some example values can be `18, 10, 15`
3. In the third case (assuming array size 20), we check that the victim transaction is after the front transaction, the victim one is after the back transaction, and the back one is before the front one. Some example values can be `10, 17, 5`

To sum up: Considering a circular structure (where the element after the last one is the first one), the transactions must be in some permutation that is in ascending order

GMX-Bots/gmx5-detect-frontruns/src/agent.ts
ericmartihaynes: I did it this way because of certain operations with the cleaner that need to occur before the victim and frontrunner are deleted from history. I tried to change it and move the cleaner to another part, but it was causing errors. I think in this case it might not be worth the time to change, but for next bots I will try to use a more clear pattern like the one you suggested

GMX-Bots/gmx5-detect-frontruns/src/agent.ts
ericmartihaynes: Causes no real problems but should not be incremented in this situation, fixed it

GMX-Bots/gmx5-detect-frontruns/src/agent.ts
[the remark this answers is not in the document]
yasmine9: What you had before is the `router` as a parameter which will be passed by value as you said. 
My suggesting is passing the `networkManager` as a parameter to `initialize` and `provideHandleTx`.

yasmine9 (commented):
LGTM overall 👍. Final notes to be approved: 

- Some unused variables still have to be removed from the bot tests file. What I could see is import `ethers` from `forta-agent` and `mockProvider` variable ( consider removing the import as well). Check if any other are not used. 

- Please update the package-lock.json. this is an old version. 

- I left a comment concerning the router address discussion.

GMX-Bots/gmx5-detect-frontruns/src/agent.ts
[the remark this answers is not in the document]
ericmartihaynes: Okay, I understand what you mean now. The issue is if I have `networkManager` as a global variable, I can't assign it its value outside `initialize` or `provideHandleTx` (because there is an async call). So then I would have to change it's value when `initialize` is run. But wouldn't I run into the same issue, where `networkManager` will be passed to `provideHandleTx` by value?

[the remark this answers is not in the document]
yasmine9: Yes, in that case you'll have the same problem. 

But you can avoid the async call when you create a new `NetworkManager` object, by passing only `util.data` as a parameter. Then initializing the chainId later by calling `init` function inside `initialize`.  You'll have something like: 

```
 const networkManager = new NetworkManager(util.data)

export const initialize = (provider: ethers.providers.Provider) => async () => {
await networkManager.init(provider)
...
}
```

yasmine9 (approved):

ericmartihaynes:
@e1Ru1o Please approve this PR when you can

GMX-Bots/gmx5-detect-frontruns/package.json
```diff
@@ -0,0 +1,40 @@
...
+  },
+  "devDependencies": {
+    "@types/jest": "^27.0.1",
+    "@types/nodemon": "^1.19.0",
+    "jest": "^27.0.6",
+    "nodemon": "^2.0.8",
+    "prettier": "^2.6.2",
+    "ts-jest": "^27.0.3",
+    "typescript": "^4.3.4",
+    "forta-agent-tools": "^2.0.7"
```
e1Ru1o, line 38: this should be in `dependencies` this package is used in the `agent.ts`

e1Ru1o (changes requested):
Ready to merge after this change

0xtaf (approved):

PR #324, opened 2022-08-23
ETH02 - Changes in Block Difficulty

Vxatz:
# Changes in Block Difficulty

## Types of changes

- [x] New Bot
- [ ] Fix a bot error

## New Bot PR readiness

- [x] Bot contains tests

## Comments (optional)

### DIFF
--- a/ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
+++ b/ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
@@ -18,7 +18,7 @@ export const provideHandleBlock = (
     const findings: Finding[] = [];
 
     // check if TTD has been reached
-    if (blockEvent.block.totalDifficulty >= TERMINAL_TOTAL_DIFFICULTY) return findings;
+    if (new BigNumber(blockEvent.block.totalDifficulty).gte(TERMINAL_TOTAL_DIFFICULTY)) return findings;
 
     // executed on the first run
     if (!init) {
--- a/ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
+++ b/ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
@@ -54,8 +54,8 @@ export const provideHandleBlock = (
     if (changePercentage.gte(threshold)) {
       findings.push(
         createFinding(
-          currentDifficulty.toString(),
-          blockDifficultyMovingAverage.toString(),
+          currentDifficulty.toString(10),
+          blockDifficultyMovingAverage.toString(10),
           changePercentage.toString(10),
           threshold.toString(10)
         )
--- a/ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
+++ b/ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
@@ -12,7 +12,6 @@ export const provideHandleBlock = (
 ): HandleBlock => {
   let blockDifficulties: BigNumber[] = new Array<BigNumber>();
   let blockDifficultyMovingAverage: BigNumber;
-  let blockDifficultiesSum: BigNumber = new BigNumber(0);
 
   return async (blockEvent: BlockEvent) => {
     const findings: Finding[] = [];
@@ -38,9 +37,7 @@ export const provideHandleBlock = (
     const currentDifficulty: BigNumber = new BigNumber(blockEvent.block.difficulty);
     blockDifficulties.push(currentDifficulty);
     //calculate the new moving average
-    for (let i = 0; i <= blockDifficulties.length - 1; i++) {
-      blockDifficultiesSum = blockDifficultiesSum.plus(blockDifficulties[i]);
-    }
+    let blockDifficultiesSum: BigNumber = blockDifficulties.reduce((acc: BigNumber, curr: BigNumber) => acc.plus(curr));
     blockDifficultyMovingAverage = blockDifficultiesSum.dividedBy(blockDifficulties.length);
     //clear the sum
     blockDifficultiesSum = new BigNumber(0);
--- a/ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
+++ b/ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
@@ -53,7 +53,8 @@ export const provideHandleBlock = (
           currentDifficulty.toString(10),
           blockDifficultyMovingAverage.toString(10),
           changePercentage.toString(10),
-          threshold.toString(10)
+          threshold.toString(10),
+          numberOfBlocksToCheck.toString()
         )
       );
     }
--- a/ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
+++ b/ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
@@ -11,7 +11,6 @@ export const provideHandleBlock = (
   threshold: BigNumber
 ): HandleBlock => {
   let blockDifficulties: BigNumber[] = new Array<BigNumber>();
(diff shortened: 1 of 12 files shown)


ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (blockEvent.block.totalDifficulty >= TERMINAL_TOTAL_DIFFICULTY) return findings;
+
+    // executed on the first run
+    if (!init) {
+      init = true;
+      //populate the blockDifficulties array in ascending block order
+      for (let i = 0; i <= numberOfBlocksToCheck - 1; i++) {
+        const blockDifficulty = new BigNumber(
+          (await provider.getBlock(blockEvent.blockNumber - numberOfBlocksToCheck + i))._difficulty.toString()
+        );
```
0xtaf, line 30: since there are 100 blocks that uses provider, would it make sense if we add `try/catch` here and if there's no error, making `init = true`?

Vxatz: I think `try/catch` should only be used in calls that can actually fail, i.e. calls to non-existing addresses or non-existing contract functions. As there are multiple nodes, I think we can, in principle, take for granted that this is going to work/the nodes use the "best" rpcUrl available, or at least that's what I've got in mind:D

0xtaf: okay because it takes around 2 minutes that's why I was worried but we can monitor the deployment :D 👍

Vxatz: Hmm, this is too much, can it be a connection issue? It's closer to 25-30" for me using the default endpoint

ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
```diff
@@ -0,0 +1,74 @@
...
+): HandleBlock => {
+  let blockDifficulties: BigNumber[] = new Array<BigNumber>();
+  let blockDifficultyMovingAverage: BigNumber;
+  let blockDifficultiesSum: BigNumber = new BigNumber(0);
+
+  return async (blockEvent: BlockEvent) => {
+    const findings: Finding[] = [];
+
+    // check if TTD has been reached
+    if (blockEvent.block.totalDifficulty >= TERMINAL_TOTAL_DIFFICULTY) return findings;
```
0xtaf, line 21: This condition will never be true because `totalDifficulty` is hex strings and `TERMINAL_TOTAL_DIFFICULTY` is number.

0xtaf (changes requested):

ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
0xtaf: oh definitely a country issue then 😄 😭

ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
Vxatz: Oops

ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    //calculate the new moving average
+    for (let i = 0; i <= newBlockDifficulties.length - 1; i++) {
+      blockDifficultiesSum = blockDifficultiesSum.plus(newBlockDifficulties[i]);
+    }
+    blockDifficultyMovingAverage = blockDifficultiesSum.dividedBy(newBlockDifficulties.length);
+
+    //clear the sum
+    blockDifficultiesSum = new BigNumber(0);
+    blockDifficulties = newBlockDifficulties;
```
0xtaf, line 49: my suggestion would be to use `shift` method of an array. it mutates the array and removes the first element. but `slice` method keeps the original array and creates a new array. so in this code it will create a new array of length 99 in each block. if `shift` was used, it would be something like this:

```suggestion
    blockDifficulties.shift();
    //add new block difficulty
    const currentDifficulty: BigNumber = new BigNumber(blockEvent.block.difficulty);
    blockDifficulties.push(currentDifficulty);

    //calculate the new moving average
    for (let i = 0; i <= blockDifficulties.length - 1; i++) {
      blockDifficultiesSum = blockDifficultiesSum.plus(blockDifficulties[i]);
    }
    blockDifficultyMovingAverage = blockDifficultiesSum.dividedBy(blockDifficulties.length);

    //clear the sum
    blockDifficultiesSum = new BigNumber(0);
```

Vxatz: Of course:D

ETH-Merge-Bots/Changes-in-Block-Difficulty/src/agent.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    //remove the oldest block difficulty
+    blockDifficulties.shift();
+    //add new block difficulty
+    const currentDifficulty: BigNumber = new BigNumber(blockEvent.block.difficulty);
+    blockDifficulties.push(currentDifficulty);
+    //calculate the new moving average
+    for (let i = 0; i <= blockDifficulties.length - 1; i++) {
+      blockDifficultiesSum = blockDifficultiesSum.plus(blockDifficulties[i]);
+    }
```
0xtaf, line 43: this is okay and fast but this can also be written with `reduce` function. just fyi.

```suggestion
    blockDifficultiesSum = blockDifficulties.reduce(
      (accumulator: BigNumber, curr: BigNumber) => accumulator.plus(curr),
      new BigNumber("0")
    );
```

Vxatz: That's way better, and I can omit the `new BigNumber("0")` as it's initialized above, right?

0xtaf: yes that's the starting initial sum, you can keep it or remove it but it's generally kept to clearly show that the initial value starts from 0

0xtaf: and when you do this, some tests fail. so you might wanna check them too

Vxatz: Can you `git pull` and check again? They seem to be working fine for me

0xtaf: oh okay now it worked 👍 btw now you can remove `blockDifficultiesSum = new BigNumber(0);` it's not necessary anymore

RCantu92 (commented):
Great job, Vasilis! Mostly minor comments.

ETH-Merge-Bots/Changes-in-Block-Difficulty/src/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+export const createFinding = (
+  blockDifficulty: string,
+  movingAverage: string,
+  changePercentage: string,
+  threshold: string
+): Finding => {
+  if (blockDifficulty > movingAverage) {
+    return Finding.fromObject({
+      name: "Unusual Block Difficulty Increase Detection",
+      description: `Block difficulty increased more than ${threshold}% compared to the moving average`,
```
RCantu92, line 18: I think we could add the amount of blocks the moving average covers to the description. Thoughts?

RCantu92 (approved):
One more suggestion. Otherwise, approved ✅

0xtaf (approved):
great job 👏

PR #338, opened 2022-09-05
UMA09 - Large Relayer Token Balance change

MakC-Ukr:
# UMA09 - Large Relayer Token Balance change

## Types of changes

- [x] New Bot
- [ ] Fix a bot error

## New Bot PR readiness

- [x] Bot contains tests


UMA-Bots/Large-Relayer-Token-Balance-Change/src/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/*
+ * @param amount: amount of tokens transferred
+ * @param token: monitored wallet address
+ * @param fundsIn: boolean value indicating whether the transfer was made in or out of monitored wallet address
+ */
+export function getFindingInstance(amount: string, addr: string, fundsIn: string) {
+  return Finding.fromObject({
+    name: "Large relayer tokens balance change",
+    description: "A large amount of funds was transferred from a specific relayer address",
```
Vxatz, line 53:
```suggestion
    description: "A large amount of funds was transferred from a monitored relayer address",
```

I think this sounds better

UMA-Bots/Large-Relayer-Token-Balance-Change/src/agent.ts
```diff
@@ -0,0 +1,61 @@
...
+      const { from, to, value } = transferEvent.args;
+      let valueBN: BigNumber = BigNumber.from(value);
+
+      if (networkManager.get("monitoredAddresses").includes(from)) {
+        let prevBalance = BigNumber.from(passedLru.get(transferEvent.address)![from]);
+        if (valueBN.gte(prevBalance.mul(alertThreshold).div(100))) {
+          findings.push(getFindingInstance(value.toString(), from, "false"));
+        }
+        let pastDict = passedLru.get(transferEvent.address);
+        pastDict![to] = prevBalance.sub(valueBN).toString();
```
Vxatz, line 41:
```suggestion
        pastDict![from] = prevBalance.sub(valueBN).toString();
```

UMA-Bots/Large-Relayer-Token-Balance-Change/src/agent.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+const MOCK_NM_DATA: Record<number, NetworkDataInterface> = {
+  0: {
+    monitoredTokens: [MONITORED_ERC20_ADDR, MONITORED_ERC20_ADDR_2],
+    monitoredAddresses: [TEST_MONITORED_ADDRESS],
+    alertThreshold: 50,
+  },
+};
+const networkManagerTest = new NetworkManager(MOCK_NM_DATA, 0);
+const testLru = new LRU<string, Dictionary<string>>({ max: 1000 }); // token address => { wallet address => balance }
+const testAlertThreshold = 50;
```
Vxatz, line 24:
```suggestion

```

UMA-Bots/Large-Relayer-Token-Balance-Change/src/agent.ts
```diff
@@ -0,0 +1,61 @@
...
+    let alertThreshold = networkManager.get("alertThreshold");
+    const findings: Finding[] = [];
+    const transferEventTxns = txEvent.filterLog(transferEvent, networkManager.get("monitoredTokens")); // Transfer event transactions for monitored addresses
+    transferEventTxns.forEach((transferEvent) => {
+      const { from, to, value } = transferEvent.args;
+      let valueBN: BigNumber = BigNumber.from(value);
+
+      if (networkManager.get("monitoredAddresses").includes(from)) {
+        let prevBalance = BigNumber.from(passedLru.get(transferEvent.address)![from]);
+        if (valueBN.gte(prevBalance.mul(alertThreshold).div(100))) {
```
Vxatz, line 37:
```suggestion
        if (valueBN.mul(100).gte(prevBalance.mul(alertThreshold))) {
```

Avoiding the division (applicable to L46 too)

UMA-Bots/Large-Relayer-Token-Balance-Change/src/agent.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+    const txEvent: TransactionEvent = new TestTransactionEvent()
+      .setFrom(RANDOM_ADDRESSES[0])
+      .addEventLog(TRANSFER_EVENT, MONITORED_ERC20_ADDR, [TEST_MONITORED_ADDRESS, RANDOM_ADDRESSES[0], "1000000"]);
+    const findings = await handleTransaction(txEvent);
+    expect(findings).toStrictEqual([testGetFindingInstance("1000000", TEST_MONITORED_ADDRESS, "false")]);
+  });
+
+  it("returns N finding only for those transactions that cross the percentage threshold (for monitored address and monitored token) when N > 1", async () => {
+    // current balance in testLru for MONITORED_ERC20_ADDR TEST_MONITORED_ADDRESS is 1500000
+    // current balance in testLru for MONITORED_ERC20_ADDR TEST_MONITORED_ADDRESS is 0
```
Vxatz, line 125:
```suggestion
```

The tests should be independent of each other, so I think you need to add a `testLru.clear()` in a `beforeEach` block. You could add a 2nd tx event where it is needed to add the balance in a previous block

Vxatz (changes requested):

UMA-Bots/Large-Relayer-Token-Balance-Change/src/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+  let monitoredTokens: string[] = networkManager.get("monitoredTokens");
+  let monitoredAddresses: string[] = networkManager.get("monitoredAddresses");
+
+  await Promise.all(
+    monitoredTokens.map(async (token) => {
+      let tokenContract = new ethers.Contract(token, ERC20_ABI, provider);
+      const tokenDict: Dictionary<string> = {};
+      await Promise.all(
+        monitoredAddresses.map(async (address) => {
+          let balance = BigNumber.from(await tokenContract.balanceOf(address));
```
0xtaf, line 36:
```suggestion
          let balance = BigNumber.from(await tokenContract.balanceOf(address, {blockTag: blockNumber}));
```
this should take `blockNumber` as a parameter. because right now `initialize` always checks the latest block balance. so the given tx wouldn’t work properly. (it still emits a finding because amount is greater than its entire balance, which doesn’t make sense)

UMA-Bots/Large-Relayer-Token-Balance-Change/src/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+          tokenDict[address] = balance.toString();
+        })
+      );
+      lru.set(token.toLowerCase(), tokenDict);
```
0xtaf, line 40: key shouldn't be just token address. it should be concatenation of two strings "tokenAddress-blockNumber", and for every block, you should call this function to update the balances again.

UMA-Bots/Large-Relayer-Token-Balance-Change/src/agent.ts
```diff
@@ -0,0 +1,61 @@
...
+  return async (txEvent: TransactionEvent) => {
+    let alertThreshold = networkManager.get("alertThreshold");
+    const findings: Finding[] = [];
+    const transferEventTxns = txEvent.filterLog(transferEvent, networkManager.get("monitoredTokens")); // Transfer event transactions for monitored addresses
+    transferEventTxns.forEach((transferEvent) => {
+      const { from, to, value } = transferEvent.args;
+      let valueBN: BigNumber = BigNumber.from(value);
+
+      if (networkManager.get("monitoredAddresses").includes(from)) {
+        let prevBalance = BigNumber.from(passedLru.get(transferEvent.address)![from]);
```
0xtaf, line 36: after changing the logic to get the balances with blockNumbers, this `prevBalance` should be caught from `loadLruCacheData` function with `blockNumber-1` parameter passed to it.

0xtaf (changes requested):
it took much time for me to understand what's going on xD
left this important logic related comments below. after this change, I'll check it thoroughly because these will affect many things

UMA-Bots/Large-Relayer-Token-Balance-Change/src/configurables.ts
```diff
@@ -0,0 +1,41 @@
+// @dev the following information is tentative
+
+// minimum percentage (%) change in token balance for a monitored address to trigger an alert
+export const MAINET_PERCENTAGE_CHANGE_THRESHOLD = 1;
+export const OPTIMISM_PERCENTAGE_CHANGE_THRESHOLD = 10;
+export const POLYGON_PERCENTAGE_CHANGE_THRESHOLD = 10;
+export const ARBITRUM_PERCENTAGE_CHANGE_THRESHOLD = 10;
+
+// List of token contract addresses that are monitored for transfers
```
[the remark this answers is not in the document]
MakC-Ukr: Added

UMA-Bots/Large-Relayer-Token-Balance-Change/src/agent.spec.ts
MakC-Ukr: @Vxatz, I am pushing a commit that solves the issue you mentioned, but it is in a bit different manner. I am simply initializing the lru cache each time in the `beforeEach` block. For the last test case (where adding some imaginary context is needed), I am setting the values in testLru to the imaginary context. Please give it a look and see if I should change it nevertheless.

MakC-Ukr:
@0xtaf, we discussed that in the end, only the `blockTag` implementation should be enough for this bot. I have implemented it in the latest commit, but even in these conditions, we have to default to some value if the INIT_BLOCK is more than 128 blocks in the past (I think this is the threshold beyond which ethers gives an error if we ask for data on the free version). As an example of a previous usage, I had used [this](https://github.com/NethermindEth/Forta-Agents/blob/26138f1d4a12a870fbd2a18ad4bdf421e4c20345/PancakeSwap-Bots/Large-LP-Deposit-Withdrawal/src/pool.fetcher.ts#L44) code for reference for adding the `try-catch` block.   

However, instead of initializing the values to 0, we will be getting the balance data in real time. If we add 0 as the default value, then in the first-ever transaction (even if very small) for a particular token-address pair, a finding will be emitted since the bot will assume that the balance changed from 0 to the new balance. To avoid this, I think it's better to load the balances and store them. For the transaction examples given in the README (for running `npm run tx`), this might not be completely fair (because we are comparing data from now to a transaction value from the past), but I am not sure how to get old balances data via `getEthersProvider.` 

Please let me know if you think this implementation makes sense.

0xtaf:
> @0xtaf, we discussed that in the end, only the `blockTag` implementation should be enough for this bot. I have implemented it in the latest commit, but even in these conditions, we have to default to some value if the INIT_BLOCK is more than 128 blocks in the past (I think this is the threshold beyond which ethers gives an error if we ask for data on the free version). As an example of a previous usage, I had used [this](https://github.com/NethermindEth/Forta-Agents/blob/26138f1d4a12a870fbd2a18ad4bdf421e4c20345/PancakeSwap-Bots/Large-LP-Deposit-Withdrawal/src/pool.fetcher.ts#L44) code for reference for adding the `try-catch` block.
> 
> However, instead of initializing the values to 0, we will be getting the balance data in real time. If we add 0 as the default value, then in the first-ever transaction (even if very small) for a particular token-address pair, a finding will be emitted since the bot will assume that the balance changed from 0 to the new balance. To avoid this, I think it's better to load the balances and store them. For the transaction examples given in the README (for running `npm run tx`), this might not be completely fair (because we are comparing data from now to a transaction value from the past), but I am not sure how to get old balances data via `getEthersProvider.`
> 
> Please let me know if you think this implementation makes sense.

hi @MakC-Ukr ,  this seems like another problem of updating the balances manually. great thinking on this work around solution. however it's wrong to set the balances inside `initialize` because we can't give a certain block number. we need to get rid of all the `NETWORK_INIT_BLOCK_NO` variables. it makes the code unnecessarily complicated and incomprehensible. inside initialize, we just need to initialize the network. and inside `provideHandleTransaction` if there's no key in cache, then we need to fetch the balance. 
see these repos as examples of how we have always used `balanceOf` functions.
https://github.com/NethermindEth/Forta-Agents/blob/main/Balancer-Bots/Large-Internal-Balance-Deposit-Withdrawal/src/agent.ts#L36
https://github.com/NethermindEth/Forta-Agents/blob/main/Balancer-Bots/Large-Swap/src/agent.ts
https://github.com/NethermindEth/Forta-Agents/blob/main/Benqi-Bots/Delegation-Monitor/src/agent.ts
https://github.com/NethermindEth/Forta-Agents/blob/main/PancakeSwap-Bots/Large-LP-Deposit-Withdrawal/src/agent.ts
https://github.com/NethermindEth/Forta-Agents/blob/main/Umee-Bots/Large-Borrow/src/agent.ts

and it's okay to not being able to fetch the old balances due to the free version limitations.

UMA-Bots/Large-Relayer-Token-Balance-Change/src/balance.fetcher.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { createAddress, MockEthersProvider } from "forta-agent-tools/lib/tests";
```
Vxatz, line 1:
```suggestion
import { MockEthersProvider } from "forta-agent-tools/lib/test";
import { createAddress } from "forta-agent-tools";
```

UMA-Bots/Large-Relayer-Token-Balance-Change/src/configurables.ts
```diff
@@ -0,0 +1,58 @@
+// @dev the following information is tentative
+
+// minimum percentage (%) change in token balance for a monitored address to trigger an alert
+export const MAINNET_PERCENTAGE_CHANGE_THRESHOLD = 1;
+export const OPTIMISM_PERCENTAGE_CHANGE_THRESHOLD = 10;
+export const POLYGON_PERCENTAGE_CHANGE_THRESHOLD = 10;
+export const ARBITRUM_PERCENTAGE_CHANGE_THRESHOLD = 10;
+
+// List of token contract addresses that are monitored for transfers
+export const MAINNET_MONITORED_TOKENS = [
```
Vxatz, line 10: If this list is arbitrary, I would suggest adding DAI, (and perhaps UMA, BAL, BOBA) tokens as well as they've got large liquidity in the `HubPool`.

UMA-Bots/Large-Relayer-Token-Balance-Change/README.md
```diff
@@ -0,0 +1,41 @@
...
+## Configuring the monitored wallet addresses list
+
+The list of monitored wallet addresses/token contract addresses/chain specific thresholds can be changed in `./src/configurables.ts`. 
+
+## Test Data
+
+The bot behaviour can be verified with the following transactions by running `npm run tx <TX_HASH>`:
+
+### Ethereum Mainnet
+- [0x234bb58e0b4cb18d1fb55a4fea208d57bc96851ebdff39317abbe8ef95aeafde](http1:/etherscan.io/tx/0x234bb58e0b4cb18d1fb55a4fea208d57bc96851ebdff39317abbe8ef95aeafde) (1 finding - the monitored address sent a large amount (more than threshold) of USDC and `Transfer` was emitted)
```
Vxatz, line 33:
```suggestion
- [0x234bb58e0b4cb18d1fb55a4fea208d57bc96851ebdff39317abbe8ef95aeafde](https:/etherscan.io/tx/0x234bb58e0b4cb18d1fb55a4fea208d57bc96851ebdff39317abbe8ef95aeafde) (1 finding - the monitored address sent a large amount (more than threshold) of USDC and `Transfer` was emitted)
```

UMA-Bots/Large-Relayer-Token-Balance-Change/src/agent.ts
```diff
@@ -0,0 +1,68 @@
...
+  };
+}
+
+export function provideHandleTransaction(
+  transferEvent: string,
+  networkManager: NetworkManager<NetworkDataInterface>,
+  balanceFetcher: BalanceFetcher
+): HandleTransaction {
+  return async (txEvent: TransactionEvent) => {
+    let alertThreshold = networkManager.get("alertThreshold");
```
Vxatz, line 25:
```suggestion
    const alertThreshold = networkManager.get("alertThreshold");
```

UMA-Bots/Large-Relayer-Token-Balance-Change/src/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/*
+ * @param amount: amount of tokens transferred
+ * @param addr: monitored wallet address
+ * @param fundsIn: boolean value indicating whether the transfer was made in or out of monitored wallet address
+ */
+export function getFindingInstance(amount: string, walletAddr: string, tokenAddr: string, fundsIn: string) {
+  return Finding.fromObject({
+    name: "Large relayer tokens balance change",
+    description: "A large amount of funds was transferred from a monitored relayer address",
```
Vxatz, line 14:
```suggestion
    description: "A monitored relayer address was involved in a large transfer of funds",
```

Rephrasing this because of the "from" in the original description that can be "to" as well

UMA-Bots/Large-Relayer-Token-Balance-Change/src/balance.fetcher.ts
```diff
@@ -0,0 +1,34 @@
...
+import { ERC20_ABI } from "./utils";
+
+export default class BalanceFetcher {
+  provider: providers.Provider;
+  private cache: LRU<string, BigNumber | string>;
+  private tokenContract: Contract;
+
+  constructor(provider: providers.Provider) {
+    this.provider = provider;
+    this.cache = new LRU<string, BigNumber | string>({
```
Vxatz, line 13:
```suggestion
  private cache: LRU<string, BigNumber>;
  private tokenContract: Contract;

  constructor(provider: providers.Provider) {
    this.provider = provider;
    this.cache = new LRU<string, BigNumber>({
```

As the `balance` is strictly typed as `BigNumber`, I think the `string` can be omitted

UMA-Bots/Large-Relayer-Token-Balance-Change/README.md
```diff
@@ -0,0 +1,41 @@
...
+  
+## Alerts
+
+- UMA-9
+  - Fired whenever a `Transfer` event is emitted with a large amount (more than the threshold) of monitored token transfer from a monitored wallet address
+  - Severity is always set to "low" 
+  - Type is always set to "info"
+  - Metadata :
+      - `amount`: amount of tokens transferred
+      - `addr` : the monitored address involved in the transfer
```
Vxatz, line 21:
```suggestion
      - `walletAddr`: the monitored relayer address involved in the transfer
      - `tokenAddr`: the address of the transferred token
```

UMA-Bots/Large-Relayer-Token-Balance-Change/README.md
```diff
@@ -0,0 +1,41 @@
...
+## Supported Chains
+- Mainnet
+- Optimism
+- Arbitrum
+- Polygon
+  
+## Alerts
+
+- UMA-9
+  - Fired whenever a `Transfer` event is emitted with a large amount (more than the threshold) of monitored token transfer from a monitored wallet address
```
Vxatz, line 16:
```suggestion
  - Fired whenever a `Transfer` event is emitted with a large amount (more than the threshold) of monitored token transferred from/to a monitored wallet address
```

UMA-Bots/Large-Relayer-Token-Balance-Change/src/utils.ts
```diff
@@ -0,0 +1,26 @@
+import { Finding, FindingSeverity, FindingType } from "forta-agent";
+
+export const TRANSFER_EVENT = "event Transfer(address indexed from, address indexed to, uint256 value)";
+export const ERC20_ABI = ["function balanceOf(address account) external view returns (uint256)"];
+
+/*
+ * @param amount: amount of tokens transferred
+ * @param addr: monitored wallet address
```
Vxatz, line 8:
```suggestion
 * @param walletAddr: the monitored relayer address involved in the transfer
 * @param tokenAddr: the address of the transferred token
```

Vxatz (changes requested):

MakC-Ukr:
The tests are passing .Thanks for reviewing @0xtaf @Vxatz

PR #354, opened 2023-01-20
POLY01 - Validator Signer ETH Balance Bot

codeWhizperer:
# POLY01 - Validator Signer ETH Balance Bot

## Types of changes

- [x] New Bot
- [ ] Fix a bot error

## New Bot PR readiness

- [x] Bot contains tests

## Comments (optional)


polygon-validator-signer-balance-bot/src/agent.ts
```diff
@@ -0,0 +1,27 @@
...
+    const formatThresholdToBigNumber = ethers.BigNumber.from(MINIMUM_THRESHOLD);
+    const accountBalance = await ethersProvider.getBalance(POLYGON_VALIDATOR_SIGNER_ADDRESS, txEvent.blockNumber - 1);
+    const formatBalance = accountBalance?.toString();
+    if (accountBalance.lt(formatThresholdToBigNumber) && !alertSent) {
+      findings.push(createUnderThresholdFinding(formatBalance));
+      alertSent = true;
+    } else if (accountBalance.gt(formatThresholdToBigNumber) && !alertSent) {
+      findings.push(createOverThresholdFinding(formatBalance));
+      alertSent = true;
+    }
```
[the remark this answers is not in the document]
Vxatz: With the example that I wrote about the new meaning of the flag, the lines can be changed to

```suggestion
    if (accountBalance.lt(formatThresholdToBigNumber) && flag.wasOverThresholdAlert) {
      findings.push(createUnderThresholdFinding(formatBalance));
      flag.wasOverThresholdAlert = false;
    } else if (accountBalance.gt(formatThresholdToBigNumber) && !flag.wasOverThresholdAlert) {
      findings.push(createOverThresholdFinding(formatBalance));
      flag.wasOverThresholdAlert = true;
    }
```

Vxatz (changes requested):
There are a few things that need to be addressed before approving, like the correct handling of the flag after each raised alert.

polygon-validator-signer-balance-bot/src/findings.ts
```diff
@@ -0,0 +1,30 @@
...
+      balance: balance,
+    },
+  });
+};
+
+export const createOverThresholdFinding = (balance: string) => {
+  return Finding.fromObject({
+    name: "Account balance greater than threshold!",
+    description: `Account balance (${balance.toString()}) above threshold (${MINIMUM_THRESHOLD})`,
+    alertId: "POLY01",
```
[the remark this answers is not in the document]
RCantu92: You are correct. We weren't requested to have a singular alert id.

polygon-validator-signer-balance-bot/src/agent.spec.ts
```diff
@@ -0,0 +1,60 @@
+import { FindingType, FindingSeverity, Finding, ethers, HandleBlock, createBlockEvent } from "forta-agent";
+import { provideHandleTransaction } from "./agent";
+import { POLYGON_VALIDATOR_SIGNER_ADDRESS } from "./constants";
+
+// let MINIMUM_THRESHOLD_BIG = ethers.BigNumber.from(MINIMUM_THRESHOLD)
+describe("POLYGON-VALIDATOR-SIGNER BOT TEST SUITE", () => {
```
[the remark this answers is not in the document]
RCantu92: To piggyback off of Vasilis comments, the additional tests that would be great to see are:
* Returns only one finding when the balance is below the threshold for more than two blocks
    * Could do this by having two different instances of passing a block to the block handler, and checking the returned amount of findings after each. The first should return one finding, the second should return nothing, despite having the same balance as the first
 * Same as above, but for balance above a threshold.
    * This would entail three blocks. The first would need to be above the threshold. The next block will need to drop below the threshold, creating a finding. Then the last block should remain below the threshold, and not create a finding.
  * Similar as the second, but the third block should come back up and pass the threshold to create a second finding.
  * The bot should not create a finding for an address that is not the Polygon Signer address
    * For this, you could pick out one of the above scenarios, but switch out the address whose balance changes

polygon-validator-signer-balance-bot/src/agent.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  beforeAll(() => {
+    handleBlock = provideHandleTransaction(mockEthersProvider);
+  });
+
+  it("returns  findings if balance is above threshold", async () => {
+    const balance = ethers.BigNumber.from("2000000000000000000");
+    const formatThresholdToBigNumber = ethers.BigNumber.from("1000000000000000000");
+    mockEthersProvider.getBalance.mockReset();
```
RCantu92, line 24: This `.getBalance()` call could be placed in a `beforeEach()` outside of an individual unit test.

polygon-validator-signer-balance-bot/src/agent.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    block: { hash: "0xa", number: 1 } as any,
+  });
+
+  beforeAll(() => {
+    handleBlock = provideHandleTransaction(mockEthersProvider);
+  });
+
+  it("returns  findings if balance is above threshold", async () => {
+    const balance = ethers.BigNumber.from("2000000000000000000");
+    const formatThresholdToBigNumber = ethers.BigNumber.from("1000000000000000000");
```
RCantu92, line 23:
```suggestion
    const balance = ethers.BigNumber.from("2000000000000000000"); // 2 ETH
    const formatThresholdToBigNumber = ethers.BigNumber.from("1000000000000000000"); // 1 ETH
```
When using large numbers like this, it helps to include a comment stating its value for legibility for a reviewer.

polygon-validator-signer-balance-bot/src/agent.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+        type: FindingType.Info,
+        protocol: "POLYGON-VALIDATOR-SIGNER-BOT",
+        metadata: {
+          balance: balance.toString(),
+        },
+      }),
+    ]);
+  });
+
+  it("returns empty array while account balance is greater than threshold", async () => {
```
RCantu92, line 48: This shouldn't rely on the state of a different unit test. If you want to run a test in which you're having the bot check two distinct blocks, you can do so in the same unit test by calling the `handleBlock()` more than once in the same unit test with different blocks, and check the return values after each.

Along the same lines, the state should be cleared in between tests, and therefore, this test should fail because it is relying on the bot having memory of the actions taken place in the previous test.

RCantu92 (changes requested):
Good progress, though there are some areas of improvement

polygon-validator-signer-balance-bot/README.md
```diff
@@ -0,0 +1,27 @@
+# POLY01 - Validator Signer ETH Balance Bot
```
RCantu92, line 1:
```suggestion
# Polygon Validator Signer ETH Balance Bot
```
Small fix

polygon-validator-signer-balance-bot/README.md
```diff
@@ -0,0 +1,27 @@
...
+   - `balance`: monitored address ETH balance
+
+## Alerts: Above Threshold
+- POLY-02
+  - Fired when the account balance is above  specified threshold
+  - Severity is always set to "info" 
+  - Type is always set to "info"
+  - Metadata fields:
+   - `balance`: monitored address ETH balance
+
```
RCantu92, line 27:
```suggestion
## Alerts
- POLY-01
  - Fired when the account balance is below  specified threshold
  - Severity is always set to "info" 
  - Type is always set to "info"
  - Metadata fields:
   - `balance`: monitored address ETH balance
- POLY-02
  - Fired when the account balance is above  specified threshold
  - Severity is always set to "info" 
  - Type is always set to "info"
  - Metadata fields:
   - `balance`: monitored address ETH balance

```
While I understand why the alerts were split up, it's not really necessary and it can be enough by having the details be underneath their respective alert id.

polygon-validator-signer-balance-bot/src/findings.ts
```diff
@@ -0,0 +1,29 @@
+import { Finding, FindingSeverity, FindingType } from "forta-agent";
+import { MINIMUM_THRESHOLD } from "./constants";
+
+export const createUnderThresholdFinding = (balance: string) => {
+  console.log(MINIMUM_THRESHOLD.toString());
```
RCantu92, line 5: Please remove this `console.log()`.

polygon-validator-signer-balance-bot/src/agent.ts
```diff
@@ -0,0 +1,40 @@
...
+      flag.wasOverThresholdAlert = false;
+    }
+  };
+};
+
+export function provideHandleBlock(ethersProvider: ethers.providers.JsonRpcProvider, flag: Flag): HandleBlock {
+  return async (txEvent: BlockEvent): Promise<Finding[]> => {
+    const findings: Finding[] = [];
+    const accountBalance = await ethersProvider.getBalance(POLYGON_VALIDATOR_SIGNER_ADDRESS, txEvent.blockNumber - 1);
+    const formatBalance = accountBalance?.toString();
```
RCantu92, line 25:
```suggestion
    const balanceString = accountBalance?.toString();
```
Suggestion for variable rename for clarity.

polygon-validator-signer-balance-bot/src/agent.spec.ts
```diff
@@ -0,0 +1,147 @@
+import { FindingType, FindingSeverity, Finding, getJsonRpcUrl, ethers, HandleBlock, Initialize } from "forta-agent";
+import { provideHandleBlock, provideInitialize } from "./agent";
+import { TestBlockEvent } from "forta-agent-tools/lib/test/index";
+import { Flag, POLYGON_VALIDATOR_SIGNER_ADDRESS, MINIMUM_THRESHOLD } from "./constants";
```
RCantu92, line 4:
```suggestion
import { Flag } from "./constants";
```
Suggestion to use mock _signed address_ and _minimum threshold_ rather than the same that are used by the bot. Therefore, these shouldn't be imported.

RCantu92 (commented):
Mostly minor comments

polygon-validator-signer-balance-bot/src/agent.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+
+describe("POLYGON-VALIDATOR-SIGNER BOT TEST SUITE", () => {
+  let handleBlock: HandleBlock;
+  let mockEthersProvider: ethers.providers.JsonRpcProvider;
+  let initialize: Initialize;
+  let flag: Flag;
+  jest.setTimeout(1000000000);
+
+  beforeEach(async () => {
+    mockEthersProvider = new ethers.providers.JsonRpcProvider(getJsonRpcUrl());
```
RCantu92, line 14: Right now, this test suite is not using a mock `provider` and thus is actually making network calls.

Please use the `MockEthersProvider` from the `forta-agent-tools` library. Section in the library can be found [here](https://github.com/NethermindEth/general-agents-module#mockethersprovider). An example usage can be found [here](https://github.com/NethermindEth/Forta-Agents/blob/main/DYDX-Bots/Large-USDC-Staking-Withdrawals/src/agent.spec.ts#L28) and [here](https://github.com/NethermindEth/Forta-Agents/blob/main/DYDX-Bots/Large-USDC-Staking-Withdrawals/src/agent.spec.ts#L37-L42).

RCantu92 (changes requested):
Please update the test suite to not set up live network provider but instead use an actual mock implementation. After updating that, I will continue my review of the test file. I didn't proceed, as this change will fundamentally change all of the subsequent unit tests.

polygon-validator-signer-balance-bot/src/agent.ts
```diff
@@ -1,27 +1,40 @@
...
 const ethersProvider = getEthersProvider();
-let alertSent: boolean = false;
+let flag = {
+  wasOverThresholdAlert: false,
+};
+
+export const provideInitialize = (ethersProvider: ethers.providers.JsonRpcProvider, flag: Flag): Initialize => {
+  return async () => {
+    const accountBalance = await ethersProvider.getBalance(POLYGON_VALIDATOR_SIGNER_ADDRESS);
+    if (accountBalance?.gt(MINIMUM_THRESHOLD)) {
```
Vxatz, line 13: `?` is not needed (it also needs to be changed in the `provideHandleBlock`)

polygon-validator-signer-balance-bot/src/findings.ts
```diff
@@ -2,13 +2,13 @@ import { Finding, FindingSeverity, FindingType } from "forta-agent";
 import { MINIMUM_THRESHOLD } from "./constants";
 
 export const createUnderThresholdFinding = (balance: string) => {
+  console.log(MINIMUM_THRESHOLD.toString());
   return Finding.fromObject({
     name: "Account balance below threshold!",
-    description: `Account balance (${balance.toString()}) below threshold (${MINIMUM_THRESHOLD})`,
-    alertId: "POLY01",
+    description: `Account balance (${balance}) below threshold (${MINIMUM_THRESHOLD.toString()})`,
```
Vxatz, line 8: Right now, `balance` and `MINUMUM_THRESHOLD` are in wei. They can be formatted, only so that they look better. You could also add the word "ETH" after both.

polygon-validator-signer-balance-bot/src/agent.ts
```diff
@@ -1,27 +1,40 @@
...
+      flag.wasOverThresholdAlert = true;
+    } else {
+      flag.wasOverThresholdAlert = false;
+    }
+  };
+};
 
-export function provideHandleTransaction(ethersProvider: ethers.providers.JsonRpcProvider): HandleBlock {
+export function provideHandleBlock(ethersProvider: ethers.providers.JsonRpcProvider, flag: Flag): HandleBlock {
   return async (txEvent: BlockEvent): Promise<Finding[]> => {
```
Vxatz, line 22: This should be a `blockEvent`.

polygon-validator-signer-balance-bot/src/agent.ts
```diff
@@ -1,27 +1,40 @@
...
     const findings: Finding[] = [];
-    const formatThresholdToBigNumber = ethers.BigNumber.from(MINIMUM_THRESHOLD);
     const accountBalance = await ethersProvider.getBalance(POLYGON_VALIDATOR_SIGNER_ADDRESS, txEvent.blockNumber - 1);
```
Vxatz, line 24: `blockEvent.blockNumber - 1`

polygon-validator-signer-balance-bot/src/agent.spec.ts
```diff
@@ -1,60 +1,147 @@
...
-    expect(mockEthersProvider.getBalance).toHaveBeenCalledTimes(1);
-    expect(mockEthersProvider.getBalance).toHaveBeenCalledWith(
-      POLYGON_VALIDATOR_SIGNER_ADDRESS,
-      blockEvent.blockNumber - 1
-    );
-    expect(findings).toStrictEqual([
+  it("should return an array of findings when the account balance is above the threshold and an over threshold alert has not been sent", async () => {
+    jest.spyOn(mockEthersProvider, "getBalance").mockResolvedValue(ethers.BigNumber.from("15000000000000000000"));
+    flag.wasOverThresholdAlert = false;
+    const txEvent = new TestBlockEvent().setNumber(10);
```
Vxatz, line 68: `blockEvent` *

polygon-validator-signer-balance-bot/src/agent.spec.ts
```diff
@@ -1,60 +1,147 @@
...
+  it("should return an empty array when the account balance is above the threshold and an over threshold alert has been sent", async () => {
+    jest.spyOn(mockEthersProvider, "getBalance").mockResolvedValue(ethers.BigNumber.from("15000000000000000000"));
+    flag.wasOverThresholdAlert = true;
+    const txEvent = new TestBlockEvent().setNumber(10);
+    const findings = await handleBlock(txEvent);
+    expect(findings).toEqual([]);
+    expect(flag.wasOverThresholdAlert).toBe(true);
+  });
+
+  it("should return an empty array when the account balance is above the threshold and an over threshold alert has been sent", async () => {
```
Vxatz, line 94: This description is currently wrong.

Vxatz (changes requested):
Great progress 👍

polygon-validator-signer-balance-bot/src/agent.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  it("throws an error when the provider fails to get the account balance", async () => {
+    mockEthersProvider.getBalance = jest.fn(() => Promise.reject("failed to fetch balance"));
+    try {
+      await initialize();
+    } catch (error) {
+      expect(error).toEqual("failed to fetch balance");
+      expect(mockEthersProvider.getBalance).toHaveBeenCalledWith(mockUser);
+    }
+  });
```
Vxatz, line 49: This test is not mimicking the code's functionality, as this error is not actually caught in the `provideInitialize`, and it thus can be removed.

polygon-validator-signer-balance-bot/src/agent.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  beforeEach(async () => {
+    mockEthersProvider = {
+      getBalance: jest.fn(),
+    };
+    flag = { wasOverThresholdAlert: false };
+
+    mockUser = createAddress("0x01");
+    initialize = provideInitialize(mockEthersProvider as any, flag, mockUser);
+    handleBlock = provideHandleBlock(mockEthersProvider as any, flag, mockUser);
```
Vxatz, line 22:
```suggestion
    initialize = provideInitialize(mockEthersProvider, flag, mockUser);
    handleBlock = provideHandleBlock(mockEthersProvider, flag, mockUser);
```

The provider is already of type `any`.

polygon-validator-signer-balance-bot/README.md
```diff
@@ -0,0 +1,27 @@
...
+- POLY-01
+  - Fired when the account balance is below  specified threshold
+  - Severity is always set to "info" 
+  - Type is always set to "info"
+  - Metadata fields:
+   - `balance`: monitored address ETH balance
+
+
+- POLY-02
+  - Fired when the account balance is above  specified threshold
```
Vxatz, line 22:
```suggestion
  - Fired when the account balance is above specified threshold
```

polygon-validator-signer-balance-bot/README.md
```diff
@@ -0,0 +1,27 @@
...
+- when account balance falls below the threshold (1 ETH)
+- when account balance recovers and raises above the threshold
+
+## Supported Chains
+- Ethereum
+
+
+## Alerts
+- POLY-01
+  - Fired when the account balance is below  specified threshold
```
Vxatz, line 14:
```suggestion
  - Fired when the account balance is below specified threshold
```

polygon-validator-signer-balance-bot/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Description
+This bot monitors the ETH balance in the account [0xa8B52F02108AA5F4B675bDcC973760022D7C6020](https://etherscan.io/address/0xa8B52F02108AA5F4B675bDcC973760022D7C6020) and triggers an alert;
+- when account balance falls below the threshold (1 ETH)
+- when account balance recovers and raises above the threshold
+
+## Supported Chains
+- Ethereum
+
+
```
Vxatz, line 11:
```suggestion

```

Vxatz (changes requested):

polygon-validator-signer-balance-bot/src/agent.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export function provideHandleBlock(
+  ethersProvider: ethers.providers.JsonRpcProvider,
+  flag: Flag,
+  address: string
+): HandleBlock {
+  return async (blockEvent: BlockEvent): Promise<Finding[]> => {
+    const findings: Finding[] = [];
+    const accountBalance = await ethersProvider.getBalance(address, blockEvent.blockNumber - 1);
+    const balanceString = ethers.utils.formatEther(accountBalance.toString());
```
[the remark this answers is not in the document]
codeWhizperer: the variable `balanceString` is being passed into the createFindings as an arguement

polygon-validator-signer-balance-bot/src/agent.spec.ts
RCantu92: @Vxatz is referring, that if you wanted to have a test for when a fetch fails, you'd have to have the `try/catch` block in bot logic rather than the unit test. Then you'd test the outcome of that fail in the unit test.

Rather than having the `try/catch` in the unit test itself. Therefore, this particular unit test is not necessary and can be removed.

polygon-validator-signer-balance-bot/src/findings.ts
```diff
@@ -0,0 +1,28 @@
+import { Finding, FindingSeverity, FindingType } from "forta-agent";
+import { MINIMUM_THRESHOLD_TO_ETHER } from "./constants";
+
+export const createUnderThresholdFinding = (balance: string) => {
+  return Finding.fromObject({
+    name: "Account balance below threshold!",
+    description: `Account balance (${`${balance} ETH`}) below threshold (${`${MINIMUM_THRESHOLD_TO_ETHER}) ETH`}`,
```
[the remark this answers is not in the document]
RCantu92: This would also need to be updated in the test suite

polygon-validator-signer-balance-bot/src/agent.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+    try {
+      await initialize();
+    } catch (error) {
+      expect(error).toEqual("failed to fetch balance");
+      expect(mockEthersProvider.getBalance).toHaveBeenCalledWith(mockUser);
+    }
+  });
+
+  it("should return an array of findings when the account balance is below the threshold and an over threshold alert has been sent", async () => {
+    jest.spyOn(mockEthersProvider, "getBalance").mockResolvedValue(ethers.BigNumber.from("5000000000"));
```
RCantu92, line 52:
```suggestion
    const balance = ethers.BigNumber.from("5000000000"); // 0.000000005 ETH
    mockEthersProvider.getBalance.mockResolvedValue(balance);
```
Suggestion to clean this up a bit, add a comment stating the amount used for `balance`. You can also increase the balance to be greater but still be less than the threshold

polygon-validator-signer-balance-bot/src/agent.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+    flag.wasOverThresholdAlert = true;
+    const blockEvent = new TestBlockEvent().setNumber(10);
+    const findings = await handleBlock(blockEvent);
+    const balance = (await mockEthersProvider.getBalance(mockUser)).toString();
```
RCantu92, line 56:
```suggestion
    const balanceString = balance.toString();
```
Would recommend to use the mock return value like this to then assign it to another variable. Since you are setting the value of `balance` yourself, you can just use that.

If do change this to be `balanceString`, that would need to be changed elsewhere.

RCantu92 (commented):
Left some more minor suggestions. However, as of now, @Vxatz would be the one to have final say as to whether the bot is completely approved.

polygon-validator-signer-balance-bot/src/agent.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+import { provideHandleBlock, provideInitialize } from "./agent";
+import { TestBlockEvent } from "forta-agent-tools/lib/test/index";
+import { createAddress } from "forta-agent-tools/lib/utils";
+import { mockFlag, MINIMUM_THRESHOLD_TO_ETHER } from "./constants";
+
+describe("POLYGON-VALIDATOR-SIGNER BOT TEST SUITE", () => {
+  let handleBlock: HandleBlock;
+  let mockEthersProvider: any;
+  let initialize: Initialize;
+  let flag: mockFlag = { wasOverThresholdAlert: false };
```
Vxatz, line 11:
```suggestion
  let mockFlag: Flag = { wasOverThresholdAlert: false };
```

What Roberto suggested is to change the name of the variable, not that of the type. (The new mock type should thus be removed from `constants.ts`)

Vxatz (changes requested):
Getting close!

Vxatz (approved):
LGTM! 🎉  Great job @codeWhizperer 👏

PR #378, opened 2023-05-08
Private Key Compromise Sharding

0xtaf:
# Private Key Compromise Sharding

## Types of changes

- [ ] New Bot
- [x] Fix a bot error

- Added Max Size check 
- Updated network calls to improve performance

### DIFF
--- a/private-key-compromise/src/agent.ts
+++ b/private-key-compromise/src/agent.ts
@@ -21,7 +21,7 @@ import { ZETTABLOCK_API_KEY } from "./keys";
 import fetch from "node-fetch";
 
 const DATABASE_URL = "https://research.forta.network/database/bot/";
-const PK_COMP_TXNS_KEY = "nm-private-key-compromise-bot-key";
+const PK_COMP_TXNS_KEY = "nm-pk-compromise-bot-key";
 const BOT_ID = "0x6ec42b92a54db0e533575e4ebda287b7d8ad628b14a2268398fd4b794074ea03";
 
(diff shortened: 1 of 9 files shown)


private-key-compromise/src/agent.ts
```diff
@@ -50,58 +67,111 @@ export const provideHandleTransaction =
...
+      if (alertedAddresses.length) {
+        for (const el of alertedAddresses) {
+          delete transferObj[el.address];
+        }
+      }
+
+      alertedAddresses = alertedAddresses.filter((address) => timestamp - address.timestamp < TIME_PERIOD);
+
+      lastBlock = blockNumber;
+      console.log(`-----Transactions processed in block ${blockNumber - 7}: ${transactionsProcessed}-----`);
```
Vxatz, line 105:
```suggestion
      console.log(`-----Transactions processed in block ${blockNumber - 7}: ${transactionsProcessed}-----`);
```
Small note: This `${blockNumber - 7}` depends on the number of shards, so as you're using different sharding for each chain you can either use a generic "last block processed" or make the "- 7" dependent on the chain

private-key-compromise/src/findings.ts
Vxatz, line 1: I think it's better to split the victims' label

`      Label.fromObject({
        entity: from.toString(),
        entityType: EntityType.Address,
        label: "Victim",
        confidence: 0.6,
        remove: false,
      }),`

Vxatz (approved):

PR #419, opened 2023-08-07
FORT12.6 - Large Profit Bot Number of Transfer Events Checker

0xtaf:
# Large Profit

## Types of changes

- [ ] New Bot
- [ ] Fix a bot error
- [x] Other (please describe):

Fixed issue https://github.com/forta-network/starter-kits/issues/227

### DIFF
--- a/large-profit-bot/package.json
+++ b/large-profit-bot/package.json
@@ -1,6 +1,6 @@
 {
   "name": "large-profit-bot",
-  "version": "0.0.4",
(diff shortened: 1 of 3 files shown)


Vxatz (approved):
- I think that the previous check of the total number of logs is now redundant, you could remove it and move the new check on that line (to avoid some provider calls that are in between the old and the new check)
- Also, please update the version

Otherwise, it looks good to me 👍

PR #431, opened 2023-08-16
FORT30 - Base Bot Implementation

RCantu92:
# Bot name here

## Types of changes

- [ ] New Bot
- [ ] Fix a bot error

## New Bot PR readiness

- [ ] The `name` and `description` fields of `package.json` describe the bot properly
- [ ] The bot README is updated
- [ ] The bot README sufficiently describes what the bot does
- [ ] All the alerts that the bot emits are listed in the README
- [ ] Code is easy to follow and/or properly commented
- [ ] Bot contains tests
- [ ] All tests pass when running `npm test`
- [ ] The networks specified in the README and the networks are listed in the `package.json` are the same
- [ ] The code does what the README describes
- [ ] `repository` is added in the `package.json`

## Comments (optional)


victim-loss-identifier/src/agent.ts
```diff
@@ -1,16 +1,255 @@
...
+  fetchNftCollectionFloorPrice: () => BigNumber
+): HandleAlert {
+  return async (alertEvent: AlertEvent): Promise<Finding[]> => {
+    const findings: Finding[] = [];
+
+    const scammerAddress = alertEvent.alert.metadata["scammer_addresses"];
+    if (!Object.keys(scammersCurrentlyMonitored).includes(scammerAddress)) {
+      scammersCurrentlyMonitored[scammerAddress] = {
+        firstAlertIdAppearance: alertEvent.alertId!,
+        mostRecentActivityByBlockNumber: alertEvent.blockNumber!,
```
Vxatz, line 182: Should this be updated even if the address is already monitored?

RCantu92: You mean that `mostRecentActivityByBlockNumber` should be updated regardless of whether it is already monitored, right?

If so, that logic to handle scammers that are already included is in the block handler. 🙂 

(If that's what you mean)

victim-loss-identifier/src/agent.ts
```diff
@@ -1,16 +1,255 @@
...
+          findings.push(
+            ...(await processFraudulentNftOrders(
+              provider,
+              scammerAddress,
+              NINETY_DAYS,
+              getErc721TransfersInvolvingScammer,
+              fetchNftCollectionFloorPrice
+            ))
+          );
+          break;
```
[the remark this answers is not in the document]
RCantu92: [This](https://www.w3schools.com/js/js_switch.asp) is how I understand `break`s to work. And since this bot will evolve to include other alert ids in the future, there will inevitably be more alert ids to check in the future.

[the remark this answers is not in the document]
Vxatz: My bad 👍

victim-loss-identifier/src/agent.ts
```diff
@@ -189,10 +180,13 @@ async function processFraudulentNftOrders(
...
 
-export function provideInitialize(provider: providers.Provider, apiKeyFetcher: () => Promise<string>): Initialize {
+export function provideInitialize(provider: providers.Provider): Initialize {
   return async () => {
+    // TODO: Add the ability to fetch the API key
+    // from the secrets.json
+
     chainId = (await provider.getNetwork()).chainId;
-    apiKey = await apiKeyFetcher();
+    apiKeys = await fetchApiKeys();
```
Vxatz, line 189: Can do it like this:

```
import { getSecrets } from "./storage";
import { apiKeys } from "./types";
.....


  apiKeys = (await getSecrets()) as apiKeys;
```

victim-loss-identifier/src/agent.ts
```diff
@@ -1,16 +1,282 @@
...
+    dataFetcher = await dataFetcherCreator(provider);
+
+    alertConfig: {
+      subscriptions: [
+        {
+          botId: SCAM_DETECTOR_BOT_ID,
+          alertIds: SCAM_DETECTOR_ALERT_IDS,
+        },
+      ];
+    }
```
Vxatz, line 207: Adding `return` and `chainId`
```suggestion
    return {
      alertConfig: {
        subscriptions: [
          {
            botId: SCAM_DETECTOR_BOT_ID,
            alertIds: SCAM_DETECTOR_ALERT_IDS,
            chainId: chainId,
          },
        ],
      },
    };
```

victim-loss-identifier/src/constants.ts
```diff
@@ -0,0 +1,62 @@
...
+  GhostMarketBSC: "0x388171F81FC91EfC7338E07E52555a90c7D87972",
+  RareBoard: "0xe0cff9e46c2b50c703a1d8c3b7438e44e4c3d474",
+  Biswap: "0x23567c7299702018b133ad63ce28685788ff3f67",
+  // OKX: "0xcce3e3f79cf9091386f84610bb06947e2fc232a3" (already listed above)
+
+  // Optimism
+  Quix: "0x998ef16ea4111094eb5ee72fc2c6f4e6e8647666",
+  // OpenSea: - same contract as on Ethereum already listed above
+  // OKX: "0xcce3e3f79cf9091386f84610bb06947e2fc232a3" (already listed above)
+};
```
Vxatz, line 62: Suggestion to add Seaport 1.5 and remove exchanges from chains other than Ethereum

```suggestion
export const EXCHANGE_CONTRACT_ADDRESSES: Record<string, string> = {
  // Ethereum
  LooksRare: "0x59728544B08AB483533076417FbBB2fD0B17CE3a", // LooksRare: Exchange
  OpenSea: "0x00000000006c3852cbEf3e08E8dF289169EdE581", // OpenSea: Seaport 1.1
  OpenSea2: "0x00000000000001ad428e4906aE43D8F9852d0dD6", // OpenSea: Seaport 1.4
  OpenSea3: "0x00000000000000ADc04C56Bf30aC9d3c0aAF14dC", // OpenSea: Seaport 1.5
  Rarible: "0x9757F2d2b135150BBeb65308D4a91804107cd8D6", // Rarible: Exchange V2
  SuperRare: "0x6D7c44773C52D396F43c2D511B81aa168E9a7a42",
  X2Y2: "0x74312363e45DCaBA76c59ec49a7Aa8A65a67EeD3",
  Blur: "0x000000000000ad05ccc4f10045630fb830b95127",
  BlurMarketplace2: "0x39da41747a83aee658334415666f3ef92dd0d541",
  BlurMarketplace3: "0xb2ecfe4e4d61f8790bbb9de2d1259b9e2410cea5",
  GhostMarketEth: "0xfB2F452639cBB0850B46b20D24DE7b0a9cCb665f",
};
```

victim-loss-identifier/src/utils/findings.ts
```diff
@@ -0,0 +1,57 @@
...
+        entity: victimAddress,
+        entityType: EntityType.Address,
+        label: "Victim Address",
+        confidence: 0, // TODO: Pass from underlying alert confidence
+        remove: false,
+      }),
+      Label.fromObject({
+        entity: contractAddress,
+        entityType: EntityType.Address,
+        label: "Stolen token Address",
```
Vxatz, line 44: Suggestion to keep the label structure same as [Ice Phishing](https://github.com/NethermindEth/forta-starter-kits/blob/main/ice-phishing/src/helper.js#L673-L675)'s (after Christian original recommendation):

```suggestion
        entity: tokenId + "," + contractAddress,
        entityType: EntityType.Address,
        label: "NFT",
```

victim-loss-identifier/src/utils/findings.ts
```diff
@@ -0,0 +1,57 @@
...
+      usd_lost: totalUsdLost.toString(),
+      erc_721_usd_lost: totalUsdLostInErc721s.toString(),
+      erc_721_lost: `name: ${name} | contract: ${contractAddress} | token id: ${tokenId} | value USD: ${usdLostOnThisToken}`,
+    },
+    labels: [
+      Label.fromObject({
+        entity: victimAddress,
+        entityType: EntityType.Address,
+        label: "Victim Address",
+        confidence: 0, // TODO: Pass from underlying alert confidence
```
Vxatz, line 38: It's [always 0.7](https://github.com/forta-network/starter-kits/blob/448d0d669c0fbcf1cbf1e12696f16ddab3476479/scam-detector-py/src/constants.py#L279), thus I believe we can hardcode it.

victim-loss-identifier/package.json
Vxatz, line 1: Suggestion on the descriptions based on the README I created (to be discussed, of course):

```
"description": "This bot identifies victims of end-user attacks and calculates their losses.",
"longDescription": "The bot's functionality revolves around monitoring alerts from the Scam Detector Feed bot. When it receives an alert, the bot extracts information related to the scammer's addresses and threat category. Subsequently, it provides answers to questions about who the victims of the scammer are and what digital assets have been transferred from these victims to the scammer.",
```

victim-loss-identifier/src/utils/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+    return latestTimestampA - latestTimestampB;
+  });
+
+  // Ensure that at least one key will be deleted in the theoretical edge case where Math.floor(sortedKeys.length / 4) equals 0
+  const indexToDelete = Math.max(1, Math.floor(sortedKeys.length / 4));
+
+  for (let i = 0; i < indexToDelete; i++) {
+    delete object[sortedKeys[i]];
+  }
+}
```
RCantu92, line 42:
```suggestion
export function cleanScammersObject(scammersObject: { [key: string]: ScammerInfo }) {
  // Sort and delete 1/4 of the keys
  const sortedKeys = Object.keys(scammersObject).sort((a, b) => {
    const latestTimestampA = scammersObject[a].mostRecentActivityByBlockNumber;
    const latestTimestampB = scammersObject[b].mostRecentActivityByBlockNumber;
    return latestTimestampA - latestTimestampB;
  });

  // Ensure that at least one key will be deleted in the theoretical edge case where Math.floor(sortedKeys.length / 4) equals 0
  const entriesToDelete = Math.max(1, Math.floor(sortedKeys.length / 4));

  for (let i = 0; i < entriesToDelete; i++) {
    delete scammersObject[sortedKeys[i]];
  }
}
```
Some suggestions on naming to make it just a bit more clear.

RCantu92 (commented):
Just some comments from my review of your updates

victim-loss-identifier/src/agent.ts
```diff
@@ -1,16 +1,171 @@
...
+  NINETY_DAYS,
+  THIRTY_DAYS_IN_SECS,
+  ONE_DAY_IN_SECS,
+  MAX_OBJECT_SIZE,
+  SCAMMERS_DB_KEY,
+  VICTIMS_DB_KEY,
+} from "./constants";
 
+let chainId: number;
+let apiKeys: apiKeys;
```
Vxatz, line 20: (this is not used)

victim-loss-identifier/src/agent.ts
```diff
@@ -1,16 +1,171 @@
...
+      let largerObjectSize = scammersObjectSize < victimsObjectSize ? victimsObjectSize : scammersObjectSize;
+
+      while (largerObjectSize > MAX_OBJECT_SIZE) {
+        console.log("Cleaning Scammers Monitored Object of size: ", scammersObjectSize);
+        cleanObject(scammersCurrentlyMonitored);
+        scammersObjectSize = Buffer.from(JSON.stringify(scammersCurrentlyMonitored)).length;
+        console.log("Scammers Monitored Object after cleaning: ", scammersObjectSize);
+
+        console.log("Cleaning Victims Scammed Object of size: ", victimsObjectSize);
+        cleanObject(victimsScammed);
```
Vxatz, line 148: I think this may not work as expected, as inside the `cleanObject` we're deleting 1/4 of the object (oldest) entries and these entries may not be the same for both objects.

I believe a better solution is to pass the largest object inside the `cleanObject` and inside that function, delete the corresponding entries from the other (not-largest) object.

WDYT?

RCantu92: Thinking about it, I'm not sure _this_ new approach would be ideal either. That's because the victim's now have their own `mostRecentActivityByBlockNumber`. So if they were scammed by a different scammer, the victim would have been active due to that.

For example, say `scammerOne` scammed `victimOne` and `scammerOne` became inactive afterwards. _Then_ `scammerTwo` scammed `victimOne` a good amount of time after `scammerOne` scammed `victimOne`.

If the time came to clean out the objects, if `scammerOne` was eligible to be removed, it would then be incorrect to remove `victimOne` because they were still "active" after having been scammed by `scammerOne` when they were scammed by `scammerTwo`.

So in the end, I don't think the entries being deleted for both _need_ to be the same. If we look at a scammer, if a specific victim was removed, since they would have already been accounted for, I don't know that it matters. The inverse is even more true because a victim could be scammed by a subsequent scammer and thus needs to be counted as part of _that_ scam.

WDYT?

Vxatz (approved):

PR #433, opened 2023-08-24
FORT30 - Helpers 

Vxatz:
# Victim/Loss Identification
## Types of changes

- [x] New Bot
- [ ] Fix a bot error


victim-loss-identifier/src/fetcher.ts
```diff
@@ -0,0 +1,263 @@
...
+
+    const ethPrice = await this.getNativeTokenPrice(timestamp);
+    const floorPriceInEth = await this.getFloorPriceInEth(nftAddress);
+
+    if (!ethPrice || !floorPriceInEth) {
+      return 0;
+    }
+
+    const floorPrice = floorPriceInEth * ethPrice;
+    this.floorPriceCache.set(timestamp, floorPrice);
```
RCantu92, line 180: This line is setting the floor price to the "key" of `timestamp`. However, setting the price in this way doesn't utilize the NFT collection address. So it runs the risk of using the floor price of one NFT collection at one timestamp for an unrelated NFT collection, if the same timestamp were to be used.

Ideally, we should incorporate the address as well. Perhaps concat'ing the timestamp and address into a single key? (If wanting to keep the key a `number`, we could convert the hexadecimal value of the address to decimal? Or convert the key to a `string` type?)

victim-loss-identifier/src/fetcher.ts
```diff
@@ -0,0 +1,263 @@
...
+      }
+    }
+    return receipt;
+  };
+
+  private getNativeTokenPrice = async (
+    timestamp: number
+  ): Promise<number | undefined> => {
+    if (this.ethPriceCache.has(timestamp)) {
+      console.log(`Using cached ETH price for timestamp: ${timestamp}`);
```
RCantu92, line 77: Should this log the price as well? Just a thought.

victim-loss-identifier/src/fetcher.ts
```diff
@@ -0,0 +1,263 @@
...
+        if (price == null) {
+          tries++; // Increment tries counter for null price
+          console.log(`Fetched price is null (attempt ${tries})`);
+        } else {
+          this.ethPriceCache.set(timestamp, price);
+          return price;
+        }
+      } catch (e) {
+        tries++;
+        console.log(`Error in fetching ETH price (attempt ${tries}):`, e);
```
RCantu92, line 100:
```suggestion
        console.log(`Error in fetching ETH price (attempt ${tries}): `, e);
```
Very small suggestion to add a space for legibility of the log

victim-loss-identifier/src/fetcher.ts
```diff
@@ -0,0 +1,263 @@
...
+        await new Promise((resolve) => setTimeout(resolve, 1000)); // wait for 1 second before retrying
+      }
+    }
+    return receipt;
+  };
+
+  getTransaction = async (txHash: string) => {
+    let receipt;
+    let tries = 0;
+    const maxTries = 3;
```
RCantu92, line 56: Small suggestion that `maxTries` could be moved one scope above, so not to repeat in each getter function. Could also be named to `MAX_TRIES` since it is a constant.

victim-loss-identifier/src/fetcher.ts
RCantu92: Same thing for L151

victim-loss-identifier/src/fetcher.ts
```diff
@@ -0,0 +1,263 @@
...
+    if (!ethPrice || !floorPriceInEth) {
+      return 0;
+    }
+
+    const floorPrice = floorPriceInEth * ethPrice;
+    this.floorPriceCache.set(timestamp, floorPrice);
+    return floorPrice;
+  };
+
+  getErc721TransfersInvolvingScammer = async (
```
RCantu92, line 184: I know I used this name as well, but for me it was a bit of a place holder. So perhaps we could go with `getScammerErc721Transfers` instead just to be a bit shorter but still very descriptive?

victim-loss-identifier/src/fetcher.ts
```diff
@@ -0,0 +1,263 @@
...
+    scammerAddress: string,
+    transferOccuranceTimeWindow: number
+  ): Promise<Erc721Transfer[]> => {
+    let erc721Transfers: Erc721Transfer[] = [];
+
+    const currentTime = new Date();
+    const timeWindowInMs = transferOccuranceTimeWindow * 24 * 60 * 60 * 1000;
```
RCantu92, line 191:
```suggestion
    const currentTime = new Date();
    const oneDayInMs = 24 * 60 * 60 * 1000;
    const timeWindowInMs = transferOccuranceTimeWindow * oneDayInMs;
```
Small suggestions to make it a bit easier to get.

RCantu92 (commented):
Mostly smaller suggestions, with one "bigger" comment regarding the lack of using the NFT collection address when caching its floor price.

Let me know if I was right or wrong in what I thought about that portion.

The rest, for the most part, LGTM 👍

victim-loss-identifier/src/fetcher.ts
Vxatz: Of course! Changed it to string type

PR #453, opened 2023-10-16
Fix Performances

0xtaf:
Large Profit & High Gas Sharding


Vxatz (approved):
@0xtaf, can you also update the version and push the package-lock? (For the large profit at least)

PR #474, opened 2024-03-12
Private Key Compromise bot: Thunder Terminal FN

RCantu92:
# Bot name here

## Types of changes

- [ ] New Bot
- [ ] Fix a bot error

## New Bot PR readiness

- [ ] The `name` and `description` fields of `package.json` describe the bot properly
- [ ] The bot README is updated
- [ ] The bot README sufficiently describes what the bot does
- [ ] All the alerts that the bot emits are listed in the README
- [ ] Code is easy to follow and/or properly commented
- [ ] Bot contains tests
- [ ] All tests pass when running `npm test`
- [ ] The networks specified in the README and the networks are listed in the `package.json` are the same
- [ ] The code does what the README describes
- [ ] `repository` is added in the `package.json`

## Comments (optional)


private-key-compromise/src/utils.ts
```diff
@@ -27,6 +33,15 @@ export type AlertedAddress = {
   timestamp: number;
 };
 
+// Left two properties `optional` for
+// backwards compatibility
+export type UpdatedAlertedAddress = {
```
0xtaf, line 38: did you leave the old `AlertedAddress` as it is on purpose? because if we're going to use this one, we can just update the old `AlertedAddress` with the new properties

private-key-compromise/src/agent.ts
```diff
@@ -5,30 +5,43 @@ import {
...
 
 import fetch from "node-fetch";
 
 const DATABASE_URL = "https://research.forta.network/database/bot/";
+// Prod keys
+// const DATABASE_OBJECT_KEYS = {
+//   transfersKey: "nm-pk-comp-bot-key-1",
+//   alertedAddressesKey: "nm-pk-comp-bot-alerted-addresses-key-1",
+//   queuedAddressesKey: "nm-pk-comp-bot-queued-addresses-key-1",
+// };
```
0xtaf, line 38:
```suggestion
const getDatabaseObjectKeys = () => {
  const isProd = process.env.NODE_ENV === "production";

  return {
    transfersKey: isProd ? "nm-pk-comp-bot-key-1" : "nm-pk-comp-bot-key-1-test",
    alertedAddressesKey: isProd
      ? "nm-pk-comp-bot-alerted-addresses-key-1"
      : "nm-pk-comp-bot-alerted-addresses-key-1-test",
    queuedAddressesKey: isProd ? "nm-pk-comp-bot-queued-addresses-key-1" : "nm-pk-comp-bot-queued-addresses-key-1-test",
  };
}

const DATABASE_OBJECT_KEYS = getDatabaseObjectKeys();
```
maybe we could use `env` variable to figure out whether it's prod or test

private-key-compromise/src/utils.ts
RCantu92: My end intention wasn't to have to `type`s, but I left both for now for the beta deployment. In the end, it was meant to have just one. Will make a note to not leave both. Thanks for pointing that out. 👍

private-key-compromise/src/agent.ts
RCantu92: It wasn't meant to be for testing, but more for the beta deployment. Similar to how you or Vasilis have been adding a random character at the end of some keys to not use the prod ones.

So not sure that we need to have different sets, one for prod and one for tests.

0xtaf: I see. if it's going to continue to be the same in prod, then it can stay like this of course 👍

0xtaf (approved):

